Issue #86, opened 2014-10-16
Provide QuickCheck input values in case of non-termination


noschinl:
When a QuickCheck test runs into a timeout, Tasty does not report the input values generated by QuickCheck for the interrupted run. These are interesting in case the test triggered a non-terminating behaviour.

QuickCheck itself can provide this information if the test is interrupted by a UserException (i.e., Ctrl-C).

osa1:
Came here to open this issue.. This feature would be really awesome.

grayjay:
Printing the replay message before running the test when --quickcheck-show-replay is true would also help with debugging non-terminating runs.

Bodigrim:
Duplicate of https://github.com/UnkindPartition/tasty/issues/359.

Bodigrim:
Closing, let's continue discussion in #359.

PR #89, opened 2014-10-23
Add soft timeouts

noschinl:
Implements #86 by adding a soft-timeout option. In contrast to the existing timeout option, a soft timeout is not asynchronous and therefore allows the test runner to intercept the timeout exception.


feuerbach:
@noschinl could you remind me on the status of this PR? I vaguely remember that it wasn't ready for some reason — is that right? Has anything changed?

noschinl:
Thanks for reminding me. As mentioned above, soft timeouts need timeouts in formatMessage. We have been using both soft-timeouts and hard-timeouts simultaneously for about three months and timeouts worked properly.

There is one open issue, namely that in combination with soft timeouts, Tasty does not necessarily recognize a timeout as a timeout, but as a failed test. I'd need to look into the details again, but the fix hinges on basvandijk/unbounded-delays#5, on which I'll need to follow up.

PR #105, opened 2015-03-21
Prevent parsing non-positive number of threads via program options

### LINKED ISSUE #104: Passing -j0 or -j-1 means test runner hangs
If you pass -j0 or -j-1 to a test run, it doesn't actually spawn any test runners and thus the test hangs.

tr00per:
Fix #104


feuerbach:
Nice, thanks!

PR #107, opened 2015-04-01
Default to using the maximum number of available cores for test execution

mlacorte:
This is particularly useful because it allows multithreading to be enabled by default in a project's Cabal file without having to guess the number of available CPU cores.


feuerbach:
Makes sense, thanks for the patch.

PR #108, opened 2015-04-02
Fix # 101 Buffer output to avoid slowdowns when printing test results

### LINKED ISSUE #101: Unbuffered output of console reporter ingredient causes noticeable slowdown on Windows
The ConsoleReporter ingredient uses NoBuffering mode when outputting test results. While it works fine on Linux, there's is noticeable slowdown when printing results of test run on Windows in cmd.exe, it's derivatives like msys and cygwin consoles. The visual effect is that output is being produced character-by-character and tests that took about 10  seconds with test-framework now take 30+ seconds to complete.

The problem is reproducible on both Windows XP and Windows 7 versions with 2014 Haskell Platform, and on a variety of terminal emulators.

The offending code is from `core/Test/Tasty/Ingredients/ConsoleReporter.hs`:

```haskell
-- | A simple console UI
consoleTestReporter :: Ingredient
consoleTestReporter =
...
  (\k -> if isTerm
    then (do hideCursor; k) `finally` showCursor
    else k) $ do

      hSetBuffering stdout NoBuffering
```

So, is there a reason to use no buffering instead of line buffering?

sergv:
Fix #101 Windows slowdown issue caused by unbuffered output.

### DIFF
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -68,9 +68,10 @@ produceOutput opts tree =
       level <- ask
 
       let
-        printTestName =
+        printTestName = do
           printf "%s%s: %s" (indent level) name
             (replicate (alignment - indentSize * level - length name) ' ')
+          hFlush stdout
 
         printTestResult result = do
           rDesc <- formatMessage $ resultDescription result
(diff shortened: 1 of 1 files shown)


feuerbach:
Looks like your patch fails to build, though.

sergv:
Well, that was embarrassing. Now i's fixed.

feuerbach:
Nice, thanks.

Issue #114, opened 2015-09-07
Known-broken tests


nomeata:
This is similar to #99, but not quite the same.

In GHC we can mark test cases as `known_broken`. This means that a failing test case does not cause the test suite to fail. If, however, the test suddenly succeeds, than *that* is a failure.

This can for example be used for test-driven development: Create the tests, mark them as `known_broken`, and whoever implements the corresponding feature will remove the marker in the process.

Possible APIs could be
    
    broken :: TestTree -> TestTree

or

    broken :: String -> TestTree -> TestTree

where the latter could include some information to be displayed along the test, e.g. a ticket number.

Such a test should be reported as `FAIL (expected)` or `FAIL (expected: …)`. If it succeeds, the reporting should be `OK (unexpected)` and it should be counted towards the total count failed tests.

nomeata:
Ah, the difference to #99 is that with ignore is that the is not run at all.

feuerbach:
How is this better than having a negation within the test itself?

E.g. if these are unit tests, then you assert that a the result of `ghc testcase.hs` has the exitcode of 1.

nomeata:
Because the test itself tell how things should be. It would be weird to write

    testCase "arith" (assert (1 + 1 /= 2))

instead of 

    broken $ testCase "arith" (assert (1 + 1 == 2))

Logically equivalent, the difference in meaning (and reporting) is huge!

Also, some tests are not easily invertable. What if I have

    testCase "arith" (1 + 1 @?= 2)

– there is no `@?/=`.

feuerbach:
Ok, I thought that you wanted to apply this to things like expected compilation errors, but apparently you want something different.

The formulation of TDD that I've come across tell that your tests should in fact fail until you've implemented the feature. Why would you want to revert that?

Are there any other applications for this apart from TDD?

nomeata:
I want them to fail, but I want it to be different from an unexpected failure. For example, if I am adding a new (still failing) test case for a yet to be developed feature, then clearly I have improved the overall situation over the previous state: The code is still the same, but the test suite is more useful. Therefore, I’d like to be able to push this change. But clearly, for someone running `cabal test` it should not complain, just as before.

Similarly for newly reported bugs: Even before a fix is present it is useful to add a test case to the test suite, but again, that should not render `cabal test` pointless.

[...]

Issue #99, opened 2015-02-06
Feature request: a way to disable a test or a group of tests a la JUnit's @Ignore annotation


manzyuk:
It would be nice to have a combinator ```ignore :: TestTree -> TestTree``` (or perhaps ```ignore :: String -> TestTree -> TestTree```) with an effect similar to that of JUnit's ```@Ignore``` annotation.

feuerbach:
Can you explain? I have no idea what `@Ignore` does.

manzyuk:
The annotation does what the name suggests: it disables a test, so that it
is not run by the test harness. Ignored tests are mentioned by the summary
report as such. In particular, they are not reported as failing if they are
broken.

This does not sound like a lot, but it can be pretty convenient when doing
major code surgeries to temporarily disable known failures. You can also
optionally specify a reason a given test is disabled.

I find this approach to be a more convenient way to disable tests than by
commenting out code: in JUnit you simply add @Ignore before a test
definition to disable it. Commenting out code is particularly tedious if
you build with -Wall -Werror and have to also comment out other definitions
that become unused or redundant imports.

feuerbach:
What about `ignore = const $ testGroup "Ignored tests" []`?

manzyuk:
This does allow one to quickly disable a test or a group of tests, but it messes up the reporting.  I'd like the ignored tests to be reported as such, not to be replaced by an empty group of tests, so that when I look at a report (generated say by a CI server), I know exactly which tests are passing, which are failing, and which are ignored and for what reason.

feuerbach:
I believe @nomeata is addressing this in his new package, see #114

nomeata:
Uploaded as http://hackage.haskell.org/package/tasty-expected-failure, contains `ignoreTest`.

PR #115, opened 2015-09-08
New functions expectFail and ignoreTest

nomeata:
based on the new function `wrapTest`, thus implementing #114 and #99.


nomeata:
After having written `wrapTest`, `expectFail` and `ignoreTest` were quite simple, I did not see the point of not including them in the module as well. Those who do want to use the feature will be happy that they are there.

nomeata:
I reordered the patches so that the (less controversial) patch is first, in case you want to pull only that.

feuerbach:
Closing, see discussion in #114.

PR #142, opened 2016-07-01
Made --quickcheck-show-replay work

NovaDenizen:
see https://github.com/feuerbach/tasty/issues/141


NovaDenizen:
Forgot to add random dependency to tasty-quickcheck.cabal

Issue #141, opened 2016-06-30
--quickcheck-show-replay True does not work 


NovaDenizen:
When I include `--quickcheck-show-replay True` in the command line of a test program I expect it to print seeds for all tests, successful and failed.  Unfortunately, this does not seem possible since `Test.QuickCheck.Test.Result`'s `Success` constructor does not make the seed available as one of its fields. 

It could be made to work if tasty-quickcheck used `Test.QuickCheck.Random.mkQCGen :: Int -> QCGen` to specify all seeds passed to QuickCheck instead of allowing it to generate its own.  I'm working on a patch that does this.

NovaDenizen:
Ok, I have a patch that works.  

QuickCheck, in it's `Failure` record of type 'Result', returns the `QCGen` and int size used for the failing test, not the QCGen used to seed the start of each test series.  QuickCheck uses an irritatingly quirky mechanism (`computesize'` in Test/QuickCheck/Test.hs) to pick the size of each individual test.

QuickCheck apparently intends for a user to take the `usedSeed` and `usedSize` from the `Failure` record and stick them into the`(QCGen, Int)` pair passed in as the `replay` field of `Args`.  This results in the first test case generated will then exactly match the test case that previously failed.  The tests that come after will be entirely different thanks to different test sizes being used.

My patch generates an int seed (from `getStdRandom`) for each test run on a property, unless an int seed is passed in via the `--quickcheck-replay` flag.  So the replay field in `Args` is always of the form `Just (mkQCGen seed, 0)`.  The 0 makes QuickCheck start with a test case of size 0, the same way it would as if `Nothing` were passed in on `replay`.  So when a seed is specified, all test cases that ran before will be run again.

As another bug, the default for `--quickplay-show-replay` was `True`, which seems wrong.  It was likely never seen as a problem because of the bug this patch fixes.  I set it to `False`.

NovaDenizen:
Created a pull request https://github.com/feuerbach/tasty/pull/142

NovaDenizen:
Whoops, I forgot to include my change to tasty-quickcheck.cabal, adding `random` as a dependency.

feuerbach:
> When I include --quickcheck-show-replay True in the command line of a test program I expect it to print seeds for all tests, successful and failed.

I think this is not what most users expect. The idea is that you only care about the seed when the test fails, so that you can reproduce the problematic case. If tests pass, most users wouldn't care about the seed. This is also the reason why the option is True by default.

[...]

PR #143, opened 2016-07-01
Fix for --quickcheck-show-replay 

NovaDenizen:
See https://github.com/feuerbach/tasty/issues/141


feuerbach:
Closing this for the reasons described in #141 (I don't want to move from TFGen back to Ints).

I filed <https://github.com/nick8325/quickcheck/issues/175>; if implemented, it'd make it possible to do this properly.

feuerbach:
This is now merged, thanks.

PR #149, opened 2016-08-31
Initial stack configuration

gliptak:


feuerbach:
Thanks, but I prefer not to track stack.yaml in git for library packages.

PR #154, opened 2016-12-20
passing down extra parent names to single test runner

haishengwu-okta:
## What changed?
- the single test runner `TreeFold.foldSingle` takes extra parameter `[TestName]` for its parent test names.

## Why?
- for `tasty-ant-xml`
- when creating an xml test result, we need specify `classname` of each test case, which mean [Full class name for the class the test method is in](https://github.com/haishengwu-okta/JUnit-Schema/blob/master/JUnit.xsd#L122-L126)
- code changes for `tasty-ant-xml`: https://github.com/haishengwu-okta/tasty-ant-xml/pull/1/files#diff-919c7cdd8a5d4c23e1e7642fa90c014f
- @ocharles 

Thanks.

### DIFF
--- a/core-tests/test.hs
+++ b/core-tests/test.hs
@@ -53,7 +53,7 @@ getTestNames :: OptionSet -> TestTree -> [String]
 getTestNames =
   foldTestTree
     trivialFold
-      { foldSingle = \_ name _ -> [name]
+      { foldSingle = \_ name _ _ -> [name]
       , foldGroup = \n l -> map ((n ++ ".") ++) l
       }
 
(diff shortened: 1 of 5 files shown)


feuerbach:
My problems with this are:

1. It is a breaking change
2. It feels too ad-hoc
3. I suspect it's unnecessary

Re the last point, it seems to me that you can pass these names through a Reader monad included in `b`, can you not?

haishengwu-okta:
@feuerbach thanks your feedback and I doubt I need use `State` monad?

`Reader` is read only but test case may have multiple level parents

```
<testsuites errors="0" failures="0" tests="1" time="0.020">
  <testsuite name="Unit Tests">
    <testsuite name="Token">
      <testsuite name="GetToken">
        <testcase name="parse success response" time="0.000" classname="Unit Tests Token GetToken">
        </testcase>
      </testsuite>
    </testsuite>
  </testsuite>
</testsuites>
```

haishengwu-okta:
@feuerbach thanks your suggestions and will give an shot.

PR #155, opened 2017-01-17
Introduce optionCLParser' and use it instead of option

liskin:
We had this function in our code but I feel it really belongs in the library
itself.


feuerbach:
Seems like an improvement!

I just find the name a bit awkward and not descriptive. How about `mkOptionCLParser`?

liskin:
Could be. In that case it would maybe make sense to add `mkFlagCLParser` taking `Mod OptionFields v` instead of `Maybe Char` too, what do you think?

feuerbach:
Agreed.

liskin:
Ok I pushed an updated branch.

PR #162, opened 2017-02-28
Mention tasty-jenkins-xml in README

liskin:
Shameless plug. :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,9 @@ Some ingredients that can enhance your test suite are:
 * [tasty-ant-xml](http://hackage.haskell.org/package/tasty-ant-xml) adds a
   possibility to write the test results in a machine-readable XML format, which
   is understood by various CI systems and IDEs
+* [tasty-jenkins-xml](https://hackage.haskell.org/package/tasty-jenkins-xml)
+  extends [tasty-ant-xml](http://hackage.haskell.org/package/tasty-ant-xml) with
+  a possibility to output to console/HTML/… and XML at the same time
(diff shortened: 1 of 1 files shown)


feuerbach:
Thanks. Have you seen this discussion? Would it solve your problem? https://github.com/feuerbach/tasty/issues/158

liskin:
Oh, no, I haven't. I wrote the code a few days before that. :-(

liskin:
Anyway, I had a quick look and it looks way simpler than my solution. The only thing that it can't do is the `--exit-success` thing (which exits successfully if the test finished and the result is correctly written to file and then relies on jenkins' junit plugin to mark the build as unstable).

feuerbach:
Hmm, I don't follow. Could you explain in a bit more detail that scenario?

liskin:
We want to distinguish between build/execution failures and just failed tests. Jenkins supports this by having different job results: FAILED and UNSTABLE. When a build script exits with a non-zero exit code, jenkins marks the build as failed. When it exits with zero and outputs a few JUnit XMLs, the JUnit Plugin can then mark the build as unstable if there are any failed tests. So we need a way to exit successfully even if some tests failed. But now that I think about it, maybe we should just catch the ExitCode exception. Why haven't I thought about all this before? :-)

feuerbach:
I see, thanks for explaining. Could you get in touch with the maintainers of tasty-ant-xml and see if they are willing to incorporate the remaining functionality in that package?

liskin:
I think I can. :-)

PR #173, opened 2017-06-19
Expose 'optionSetToArgs' in tasty-quickcheck (fixes #172)

### LINKED ISSUE #172: Tasty -> QuickCheck Args
Hello!

Would it be possible to expose a function with the type

```haskell
Tasty.OptionSet -> QuickCheck.Args
```

in the `tasty-quickcheck` package? My use case is simply to keep `tasty-hspec` in sync with the behavior of `tasty-quickcheck` since `hspec` also accepts QuickCheck `Args`.

I think it would be surprising if, depending on which package you pass your tasty `OptionSet` to, the underlying QuickCheck args are actually set to one thing or another.

https://github.com/mitchellwrosen/tasty-hspec/pull/14

Thank you!

mitchellwrosen:
This patch is wanted for `tasty-hspec`, see #172.

I only bumped the patch as this function is part of the internals, I think. Thanks!


feuerbach:
Thanks, released as 0.9.1. (Since there are other packages depending on this API now, I don't think it can be considered internal.)

PR #186, opened 2017-10-31
Revise console output API.

merijn:
Make it easier to access the name of a test/group for things like labelling.

Add a way to report statistics without timing.

Updated @since annotations to reflect the current type and updated the version in accordance with PVP.

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changes
 =======
 
+Version 0.11.3
+--------------
+
+Expose and document several of the internals of
+`Test.Tasty.Ingredients.ConsoleReporter`.
+
 Version 0.11.2.5
 ----------------
 


feuerbach:
It's a shame that we need to bump the major version — it'll affect a lot of people without a good reason. But I guess that's the price of the PVP.

Could you squash the two commits together and force-push? Also, could you add a changelog entry?

merijn:
Yeah, it's a shame I didn't realise I needed this before pushing the initial version. I doubt anyone but me has actually started using this code since it was released, so not bumping the major version is *probably* fine, but I decided to err on the side of caution.

merijn:
Ok, pushed a squashed version.

feuerbach:
Cool, could you also add a changelog entry?

merijn:
Apparently I can't "git" well, so this got closed when I was trying to revise it to include the CHANGELOG. The fixed version is in #188

PR #197, opened 2017-12-25
Fixed warnings on GHC 8.4.1-alpha1

asr:
The warnings were like this one:

```
Test/Tasty/Ingredients/ConsoleReporter.hs:536:10: warning: [-Wmissing-methods]
    • No explicit implementation for
        ‘<>’
    • In the instance declaration for ‘Semigroup (Maximum a)’
```

I tested the PR with GHC 8.4.1-alpha1, 8.2.2, 8.0.2 and 7.10.3.


core/tasty.cabal
```diff
@@ -64,6 +64,10 @@ library
     -- for GHC.Generics
     build-depends: ghc-prim
 
+  if impl(ghc < 8.0)
+    -- provide/emulate `Data.Semigroups` API for pre-GHC-8
+    build-depends: semigroups == 0.18.*
+
```
feuerbach: They way I did it (with CPP pragmas) was intentional — I'd rather not add extra dependencies without a good reason.

asr: I guess I could avoid the new dependency defining `(<>) = mappend` but I prefer the current approach, that is, the GHC 8.4.1-alpha approach, because it follows the mathematical definitions.

Issue #196, opened 2017-12-24
Build error with GHC 8.4.1-alpha1


asr:
GHC 8.4.1-alpha1 was [announced](https://mail.haskell.org/pipermail/ghc-devs/2017-December/015235.html). While testing [Agda](https://github.com/agda/agda) with this version of GHC, I got various errors on tasty-silver (master branch) related to missing instances of `Semigroup` like this:

```
$ cabal install
...
Test/Tasty/Options.hs:81:10: error:
    • No instance for (Semigroup OptionSet)
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘Monoid OptionSet’
   |
81 | instance Monoid OptionSet where
   |          ^^^^^^^^^^^^^^^^
Failed to install tasty-0.12
```

These errors are reported as warnings in GHC 8.0.2/8.2.2 using the `-Wsemigroup` flag:

```
$ cabal install --force-reinstalls --ghc-options='-Wsemigroup -Werror'
```

So, fixing these warnings on GHC 8.0.2/8.2.2 fix the errors on GHC 8.4.1-alpha1.

Blocking https://github.com/agda/agda/issues/2878.

feuerbach:
Thanks for the report. It should fixed on master now — could you try it with ghc 8.4? If it works, I'll make a release.

asr:
Thanks for the fix!

Testing with GHC 8.4.1-alpha1 I got various warnings like this:

```
Test/Tasty/Options.hs:88:10: warning: [-Wmissing-methods]
    • No explicit implementation for
        ‘<>’
    • In the instance declaration for ‘Semigroup OptionSet’
```
I will create a PR fixing these warnings.

For knowing the warnings in GHC 8.0.2/8.2.2, I realised it's better to use the `-Wcompat` flag (sorry about that).

asr:
> I will create a PR fixing these warnings.

Done. See https://github.com/feuerbach/tasty/pull/197

asr:
> could you try it with ghc 8.4? If it works, I'll make a release.

Modulo some dependencies (but `tasty.cabal` is ready), I could compile with GHC 8.4.1-alpha1.

feuerbach:
Wait, that's weird. Do you know why these warnings appear?

I relied on the default implementation of `<>` (https://hackage.haskell.org/package/base-4.10.1.0/docs/Data-Semigroup.html) and indeed got no such warnings under earlier ghcs.

Have they removed the default method in 8.4?

[...]

PR #207, opened 2018-01-18
Refactoring for #196

asr:
From our previous discussion in https://github.com/feuerbach/tasty/issues/196#issuecomment-353880917, we know that `mappend` is defined as `(<>)` by default. In this PR I used this fact for refactoring the fix of #196.


feuerbach:
What's the purpose of this? It looks like it complicates things for no reason.

quasicomputational:
Soon GHC will be warning about non-canonical `Monoid` definitions (where the canonical definition is `mappend = (<>)`). There's a long-term idea of removing `mappend` from `Monoid` altogether and leaving it as an alias for `(<>)`, but I don't know if that's going to realistically happen. The general idea is that `mappend` and `(<>)` should be identical, and, since `Semigroup` is a superclass of `Monoid`, it makes the most sense to define `(<>)` explicitly.

This changeset is probably over-complicated: I'd drop the CPP and add a dependency on the semigroups package (which is already transitively pulled in through optparse-applicative), and then define `(<>)` as the source of truth and leave in `mappend = (<>)` for compatibility.

asr:
> This changeset is probably over-complicated: I'd drop the CPP and add a dependency on the semigroups package (which is already transitively pulled in through optparse-applicative), and then define `(<>)` as the source of truth and leave in `mappend = (<>)` for compatibility.

I agree. On fact, this is what I made in #197 (but you rejected it due to the dependency on semigroups).

asr:
> Soon GHC will be warning about non-canonical Monoid definitions (where the canonical definition is `mappend = (<>)`).

Note that this PR uses the canonical Monoid definition.

feuerbach:
> which is already transitively pulled in through optparse-applicative

This is a good point I haven't considered before, and when this becomes an issue, I'll do as you say. But for now, I honestly don't see why we are wasting each other's time having this discussion.

So I'll repeat the question: does this PR solve any problem right now?

asr:
> So I'll repeat the question: does this PR solve any problem right now?

No. I thought that the fix of #196 using something like

```
instance Semigroup Foo where
  (<>) = mappend
#endif
```
wasn't the best. This PR was my attempt of a better fix. Since you are thinking of including a dependency in the semigroup package, this PR is "outdated".

PR #212, opened 2018-02-12
Depend on semigroups

gwils:
so that we can offer Semigroup instances before Semigroup was in base


feuerbach:
See the previous discussion at #207.

As it happens, I am moving in the opposite direction — dropping the dependency on optparse-applicative so that tasty no longer indirectly depends on semigroups, which is itself rather fat in terms of dependencies.

So you'll need to make a strong case for why these instances are needed.

gwils:
Tasty supports old GHC versions quite far back. But these Semigroup instances are missing for anyone on GHC 7. So if I'm a user of tasty, and I intend to also support GHC 7, I will not be able to use these Semigroup instances at all.
Moreover, I will have to be aware of this and carefully program around it. I would reach for Semigroup, have it work locally, and then find out from travis that my GHC 7 builds are broken. This is not a good experience. If tasty supports older GHC versions, please support them properly by using compatibility shims like `semigroups`.

As for dependencies, even without `optparse-applicative`, we still don't pick up many from `semigroups`.

We already depend directly on:

* base
* containers
* deepseq
* ghc-prim
* tagged

mtl brings in transformers

async brings in:

* hashable, and therefore bytestring
* text, and therefore binary

So `semigroups` is only bringing in `bytestring-builder`, `nats`, `transformers-compat`, and `unordered-containers`.

`bytestring-builder` and `transformers-compat` will disable themselves on any modern `bytestring` and `transformers` version. `nats` will disable itself on a modern base (as will `semigroups`, by the way).

So in practice, semigroups only comes with one new dependency: `unordered-containers`, [which is ubiquitous anyway](http://packdeps.haskellers.com/reverse/unordered-containers)

This is a tiny cost to pay, and this change will benefit those who use tasty and support GHC 7

feuerbach:
> So if I'm a user of tasty, and I intend to also support GHC 7, I will not be able to use these Semigroup instances at all.

Give me a specific example of a Semigroup instance that you as a tasty user might plausibly need.

gwils:
Perhaps two values of type `OptionSet` will need to be combined `o1 <> o2`

I am not some theoretical library user whose use case is 100% of the library surface. I have not hit this problem personally. But it is part of the library, so it could be hit by someone with different needs.

My view is that an open source library should try to anticipate problems users might have, and fix them proactively as best you can. That's what I've tried to do here. Don't wait until somebody hits a bug before fixing it.
It seems sensible to me that if you support both GHC 7.4 and GHC 8.2, and you have a Semigroup instance on your data type, that it would work on both.

feuerbach:
> Perhaps two values of type OptionSet will need to be combined o1 <> o2

As they can, because `<>` is a synonym for `mappend` in older ghcs.

PR #215, opened 2018-04-01
Removed unused extensions

Anabra:
Removed most unused extensions from every module. This has been done using the OrganizeExtensions refactoring of the Haskell-Tools framework. See http://www.haskelltools.org/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.stack-work
 .cabal-sandbox
 cabal.sandbox.config
 cabal.project.local
--- a/core-tests/exit-status-test.hs
+++ b/core-tests/exit-status-test.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveDataTypeable, BangPatterns #-}
+{-# LANGUAGE DeriveDataTypeable #-}
 import Test.Tasty
 import Test.Tasty.Options
 import Test.Tasty.HUnit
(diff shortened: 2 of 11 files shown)


Anabra:
What is the reason for closing this PR immediately?

feuerbach:
Sorry, I didn't give an explanation because I suspected you were a bot.

The reason I closed it was:

1. It was not soliving any real problem I had.
2. It used a different style of pragmas than I use (I use a single pragma and comma-delimited extensions).
3. In one case it removed a meaningful comment.

If you genuinely want to help with the language pragmas, please move them from source files to the respective cabal files (default-extensions). Please format them one per line and preserve the comment(s).

Anabra:
I don't understand your last point. Why would I move module-specific extensions to cabal files? As far as I know, the default-extensions tag is used for globally enabled language extensions.

I'll manually reformat the pragmas and readd all necessary extensions for backward compatibility with older GHC versions.

feuerbach:
> Why would I move module-specific extensions to cabal files? As far as I know, the default-extensions tag is used for globally enabled language extensions.

Correct. Instead of maintaining per-module extension list, I'd rather maintain per-package extension lists.

Anabra:
I committed the updated version with the local changes. Since the PR is closed, it won't appear here. If you would like to review it, you can find it on the forked master branch. If you are interested, we can merge that commit.

The purpose of this PR would have been to test the newly implemented dead extension elimination feature of our tool. The per-package extension management that you mentioned is an interesting concept, we will look into it. For now, the refactoring works on a per-module basis.

feuerbach:
> The purpose of this PR would have been to test the newly implemented dead extension elimination feature of our tool.

I understand your motivation, but for me this is nuisance. It's still not good enough to be readily merged (e.g. the pragmas in `core/Test/Tasty/Core.hs` are still split, but don't bother to fix that), and the utility of the whole thing is close enough to 0 to justify spending any more time on this.

PR #217, opened 2018-05-01
Add tasty directories trick to Press

nmattia:
@feuerbach I'm regularly surprised by how flexible tasty is, I love it. Last week I needed to read the test names from within the tests and came up with a nice solution.

Do you think this is out of scope for the "Press" section?


feuerbach:
That is exactly what the press section is for — thanks for blogging about tasty!

Issue #111, opened 2015-05-19
Alignment breaks with wide Asian characters


feuerbach:
Originally [reported](https://groups.google.com/forum/#!topic/haskell-tasty/RsUDVOwWc6E) by Bongseo Jang on the mailing list.

The problem is that some characters (like [U+C774](https://codepoints.net/U+C774)) take double space in a monospaced font.

I suspect this is determined by the [EAST ASIAN WIDTH](http://www.unicode.org/reports/tr11/) property, but this needs to be confirmed yet.

feuerbach:
See also: https://en.wikipedia.org/wiki/Halfwidth_and_fullwidth_forms

feuerbach:
> Except for a few characters, which are explicitly called out as fullwidth or halfwidth in the Unicode Standard, characters are not duplicated based on distinction in width. Some characters, such as the ideographs, are always wide; others are always narrow; and some can be narrow or wide, depending on the context. The Unicode character property East_Asian_Width provides a default classification of characters, which an implementation can use to decide at runtime whether to treat a character as narrow or wide.

This isn't helping :/

tosoonee:
Why do you think it's not helping? I'm new to haskell and just want to know :-)

It seems that the width can be determined with the property East_Asian_Width, can't it? For ambiwidth, there is 'set ambiwidth=[single|double]' in vim and seems decent decision.

feuerbach:
I was specifically referring to these parts:

> some can be narrow or wide

and

> an implementation can use to decide at runtime whether to treat a character as narrow or wide

So I don't see how I can determine how much space these characters will take on a terminal.

feuerbach:
This seems to be difficult to fix and relatively low impact, so closing. If anyone comes up with a simple self-contained solution, I'd definitely consider it.

PR #223, opened 2018-06-14
Better Unicode support in status-column alignment.

leftaroundabout:
`buildTestOutput` aligns the column of `[OK]` or `[FAIL]` messages neatly along
a vertical line, to ensure readable output. It does so by measuring the
width of the name-strings, as of now, using the simple `Prelude.length`
function.

But even with a monospaced font, Unicode text generally can't be relied upon to
have a width proportional to the count of code points. Notable, CJK characters
usually need twice the width of Western characters. As a result, Tasty may
produce output like this:

tasty-bad-align

At least under Unix, the `wcwidth` function can be used to learn, for most of
Unicode, how wide the individual `Char` code points will come out on the
terminal, thus we can more reliably estimate the width and improve the
above to

tasty-good-align


leftaroundabout:
The [build failure in GHC-7.4](https://travis-ci.org/feuerbach/tasty/jobs/392242831) is apparently a GHC bug.
```
ghc: panic! (the 'impossible' happened)
  (GHC version 7.4.2 for x86_64-unknown-linux):
	cgLookupPanic (probably invalid Core; try -dcore-lint)
    ( ds1{v sepm} [lid] :: (# ( a{tv adLO} [sk] :: ghc-prim:GHC.Prim.*{(w) tc 34d} ) #) )
    static binds for:
    ( primitive-0.6.4.0:Data.Primitive.Array.array#{v rdp4} [gid[[RecSel]]] :: forall ( a{tv adpN} [tv] :: ghc-prim:GHC.Prim.*{(w) tc 34d} ).
```
Would it make sense to move the 7.4 build to _allowed failures_ on Travis?

feuerbach:
Ah awesome, this was previously reported as #111.

> Would it make sense to move the 7.4 build to allowed failures on Travis?

Is that failure related to wcwidth? I don't understand the connection — seems like it arises when compiling `primitive`.

If it is in fact caused by the new dependency, we should restrict the dependency to the newer ghcs so the library remains usable with the old ones.

Also, could you add a haddock comment to stringLength explaining why it's needed and why it's conditional on the OS and ghc version?

Would `stringWidth` be a better name for that function?

leftaroundabout:
Yeah, I can't see what the 7.4 failure could have to do with `wcwidth`; that has to be some change in `primitive-0.6.4` that triggers the bug in GHC-7.4 and 7.6 when compiling in a certain order. I recall seeing similar panics  7.4 and 7.6 with no determinable reason. Have you tried re-triggering a build that _doesn't_ depend on `wcwidth` yet, since may 30? I'd suspect it would fail as well.

About the name of that function – well, `stringWidth` would kind of suggest it computes the width of a box containing text (possibly with newlines). While that might actually make sense (introducing line breaks for over-long test names), it's not what the function currently does, so, I'd say `stringLength` is rather better. But I can see why you might find `stringWidth` preferrable. I don't really have an opinion on this.

feuerbach:
[You're right re primitive](https://travis-ci.org/feuerbach/tasty/builds/379594979); I'll look into that.

String length is usually defined as the number of characters/symbols in the string, whereas your function takes into account the graphical representation of those characters, so I think width would be a better fit here.

feuerbach:
Filed https://github.com/haskell/primitive/issues/177

feuerbach:
I pushed a workaround to master — could you rebase on top of it?

leftaroundabout:
Ah, the joy of continuous integration...

So, https://travis-ci.org/feuerbach/tasty/jobs/392681943, https://travis-ci.org/feuerbach/tasty/jobs/392681944 and https://travis-ci.org/feuerbach/tasty/jobs/392681946 seem to be Travis/Debian issues now.
```
$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install cabal-install-1.16 ghc-7.4.2
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package cabal-install-1.16
E: Couldn't find any package by glob 'cabal-install-1.16'
E: Couldn't find any package by regex 'cabal-install-1.16'
E: Unable to locate package ghc-7.4.2
E: Couldn't find any package by glob 'ghc-7.4.2'
E: Couldn't find any package by regex 'ghc-7.4.2'
apt-get.diagnostics
apt-get install failed
W: Some index files failed to download. They have been ignored, or old ones used instead.
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install cabal-install-1.16 ghc-7.4.2" failed and exited with 100 during .
```
The error in https://travis-ci.org/feuerbach/tasty/jobs/392681948 I don't understand at all.

PR #224, opened 2018-07-05
build fix for ghc-8.6

tmcdonell:


feuerbach:
Could you give some details about what problem this is fixing?

tmcdonell:
The first ghc-8.6 release candidate was released a few days ago:

https://mail.haskell.org/pipermail/ghc-devs/2018-June/015970.html

Tasty does not compile with ghc-8.6 because `Ap` is now defined for you. This patch simply uses the built-in version for base-4.12 and beyond.

feuerbach:
Do you need a release for this?

tmcdonell:
Not immediately. I think, it would be fine to wait for the official 8.6 release.

PR #229, opened 2018-10-09
GHCJS compatibility

Shimuuar:
If package is build on linux `wcwidth` & `unix` packages are added
as dependencies. `wcwidth` calls C functions which makes it not
possible to run tests with GHCJS

### DIFF
--- a/core/tasty.cabal
+++ b/core/tasty.cabal
@@ -77,7 +77,7 @@ library
     -- for GHC.Generics
     build-depends: ghc-prim
 
-  if !os(windows)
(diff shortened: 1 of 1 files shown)


feuerbach:
Is this enough to make it work? Wouldn't the build fail when importing modules from those packages?

Shimuuar:
I'm not sure whether it fixed _all_ problems  with GHCJS compatibility but tasty-hunit based tests worked without any problems since windows fallback doesn't use anything fancy

feuerbach:
Released as tasty-1.1.0.4.

Issue #231, opened 2018-12-02
localOption (HideSuccesses True)   doesn't appear to work


dustin:
I've got a test that has about 4,000 cases I'm trying to fix.  It's a generated subsuite and I'd really like to not see it unless something is broken.

`-ta --hide-successes` works but just prefixing this particular bit with `localOption (HideSuccesses True)` doesn't.

e.g.:

    testPartialTable :: [TestTree]
    testPartialTable = undefined

    [...]

    localOption (HideSuccesses True) $ testGroup "partial table" testPartialTable,

feuerbach:
Yes, HideSuccesses is a global option at the moment. It's possible to make it local, though that'd be a bit involved. Feel free to give it a try, the relevant code is in core/Test/Tasty/Ingredients/ConsoleReporter.hs.

quchen:
Just stumbled upon this, the option still does not work (tasty 1.4.2.1).

PR #232, opened 2018-12-13
Support local options in consoleOutput

dustin:
In particular, this addresses #231 by allowing HideSuccesses to be honored at result reporting time.

### DIFF
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -71,6 +71,7 @@ data TestOutput
       {- test name         -} String
       {- print test name   -} (IO ())
       {- print test result -} (Result -> IO ())
+      {- get test options -} OptionSet
       -- ^ Name of a test, an action that prints the test name, and an action
       -- that renders the result of the action.
   | PrintHeading String (IO ()) TestOutput
@@ -104,7 +105,7 @@ buildTestOutput opts tree =
     runSingleTest
       :: (IsTest t, ?colors :: Bool)
       => OptionSet -> TestName -> t -> Ap (Reader Level) TestOutput
-    runSingleTest _opts name _test = Ap $ do
+    runSingleTest opts' name _test = Ap $ do
       level <- ask
 
       let
@@ -134,7 +135,7 @@ buildTestOutput opts tree =
             (if resultSuccessful result then infoOk else infoFail) $
               printf "%s%s\n" (indent $ level + 1) (formatDesc (level+1) rDesc)
 
-      return $ PrintTest name printTestName printTestResult
+      return $ PrintTest name printTestName printTestResult opts'
 
     runGroup :: TestName -> Ap (Reader Level) TestOutput -> Ap (Reader Level) TestOutput
     runGroup name grp = Ap $ do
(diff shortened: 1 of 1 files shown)


dustin:
This was a first pass I did last night (before AoC took over).  It seemed to do what I wanted quite well.

It might be nice to go further and have only one console output mechanism, but I don't know that I understand all of your use cases.

dustin:
I don't understand that appveyor failure, and I haven't figured out how to run your tests properly (I found the built executables and everything passes).

dustin:
Optionally added a commit to remove `consoleOutputHidingSuccesses` which is now redundant.

dustin:
Removing `consoleOutputHidingSuccesses` does reveal a bit more complication since the groups don't receive options either, and that stems from core.  I'll await guidance.

core/Test/Tasty/Ingredients/ConsoleReporter.hs
```diff
@@ -71,6 +71,7 @@ data TestOutput
       {- test name         -} String
       {- print test name   -} (IO ())
       {- print test result -} (Result -> IO ())
+      {- get test options -} OptionSet
```
feuerbach: Could you align this with the other comments please? :)

core/Test/Tasty/Ingredients/ConsoleReporter.hs
```diff
@@ -194,46 +195,27 @@ consoleOutput :: (?colors :: Bool) => TestOutput -> StatusMap -> IO ()
...
-    foldTest _name printName getResult printResult =
+    foldTest _name printName getResult printResult opts =
       ( Traversal $ do
-          printName :: IO ()
+          let HideSuccesses hideSuccesses = lookupOption opts
           r <- getResult
-          printResult r
+          _ <- if resultSuccessful r && hideSuccesses
+               then do pure $ Any False
+               else do printName :: IO (); printResult r :: IO (); return $ Any True
```
feuerbach: Any True and Any False don't do anything since the result is ignored. This hints at an issue: you are not erasing the empty headings like the original implementation did. Try compiling `core-tests/hide-successes-test.hs` on master and your branch and compare the output.

core/Test/Tasty/Ingredients/ConsoleReporter.hs
```diff
@@ -194,46 +195,27 @@ consoleOutput :: (?colors :: Bool) => TestOutput -> StatusMap -> IO ()
...
-    clearAboveLine = do cursorUpLine 1; clearThisLine
-    clearThisLine = do clearLine; setCursorColumn 0
-
 streamOutputHidingSuccesses :: (?colors :: Bool) => TestOutput -> StatusMap -> IO ()
 streamOutputHidingSuccesses toutput smap =
   void . flip evalStateT [] . getApp $
     foldTestOutput foldTest foldHeading toutput smap
   where
-    foldTest _name printName getResult printResult =
+    foldTest _name printName getResult printResult _opts =
```
feuerbach: I think you should update this function too; otherwise your new feature will work only in terminals.

dustin:
I'm going to go ahead and close this.  I don't think it's actually that important as I've found better ways to accomplish my goals.

PR #233, opened 2018-12-21
color=always in Emacs

fommil:
As of https://github.com/haskell/haskell-mode/pull/1608 (and https://github.com/haskell/haskell-mode/pull/1609) Emacs is able to handle colour output in tasty tests by default.

Before I progress with this PR, I wanted to check if it's something you'd be interested in accepting. It saves me having to add `--color=always` to every invocation.

It is relatively easy to detect that we are within Emacs, we just look for the envvar `INSIDE_EMACS`. The logic to match how this is done in the JVM world is at https://github.com/jline/jline2/pull/288 / https://github.com/jline/jline2/commit/5366158937c49b6e277345e3e543ba61b6fe33b4

TL;DR emacs supports ansi colours but not echo and navigation, when compared to other "smart" terminals.

The situation I'd be concerned about is when cabal runs multiple tasty jobs in a batch and puts their output into a file instead of the console. In such cases I don't think we want colour output... but we would have the same problem in other terminals also. Is there a way to detect that?

A user or project level `.tasty` containing some options would be another way to do this.

### DIFF
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -398,6 +398,7 @@ consoleTestReporter =
 
       do
       isTerm <- hSupportsANSI stdout
+      isTermColor <- hSupportsANSIColor stdout
 
       (\k -> if isTerm
         then (do hideCursor; k) `finally` showCursor
@@ -406,7 +407,7 @@ consoleTestReporter =
           hSetBuffering stdout LineBuffering
 
           let
-            ?colors = useColor whenColor isTerm
+            ?colors = useColor whenColor isTermColor
 
           let
             toutput = buildTestOutput opts tree
--- a/core/tasty.cabal
+++ b/core/tasty.cabal
@@ -66,7 +66,7 @@ library
     optparse-applicative >= 0.14,
     unbounded-delays >= 0.1,
     async >= 2.0,
-    ansi-terminal >= 0.6.2
+    ansi-terminal >= 0.9
 
   if flag(clock)
     build-depends: clock >= 0.4.4.0
--- a/stack.yaml
+++ b/stack.yaml
@@ -7,6 +7,7 @@ packages:
 - core-tests/
 extra-deps:
 - wcwidth-0.0.2
+- ansi-terminal-0.9
 # this is needed for travis, because call-stack is absent from the old snapshots
 - call-stack-0.1.0
 flags: {}


feuerbach:
> TL;DR emacs supports ansi colours but not echo, when compared to other "smart" terminals.

Could you elaborate on what "but not echo" means?

feuerbach:
I'm all for solving the issue that you have with Emacs, but I'm not very comfortable with adding Emacs-specific code to tasty—and this has almost nothing to do with the fact that I'm a vim user ;-)

I feel like the right place to add Emacs-specific checks may be in ansi-terminal, or maybe try to make use of terminfo (but would that work on Windows?). Then we can adjust tasty w.r.t. more abstract capabilities that a terminal provides, instead of checking for Emacs specifically.

fommil:
I don't think echo is relevant for tasty. It's when the user inputs commands, echo would then add them to stdout.

Unfortunately it's an all-or-nothing check in `ansi-terminal` http://hackage.haskell.org/package/ansi-terminal-0.8.2/docs/System-Console-ANSI.html#g:11 whereas Emacs can handle colours but not necessarily the interactive features like cursor navigation. What we need is a `hSupportsANSIColor` and Emacs can return `True`. Would you be open to a PR adding that kind of more refined check?

feuerbach:
> Would you be open to a PR adding that kind of more refined check (to ansi-terminal)?

Yes, I think this would be the right approach.

fommil:
@feuerbach ok, I sent a PR to the upstream lib.

fommil:
To get this to work while the upstream is unreleased add this to `cabal.project`

```
source-repository-package
    type: git
    location: https://github.com/fommil/ansi-terminal.git
    tag: d25db05b900db889dddc91e93cea47c79621225a
```

fommil:
@feuerbach now that `ansi-terminal` 0.9 is released, I updated my PR :smile:

feuerbach:
But isTerm isn't just used to output colors; it's also used to erase stuff. So you may need to distinguish those two capabilities.

fommil:
ok, I'll take a look at that soon.

fommil:
@feuerbach like this?

fommil:
`--hide-successes` works fine in Emacs, and I'm seeing colours.

I pushed an update to the `stack.yaml` so the windows build will hopefully go green.

feuerbach:
Oops, looks like I broke the CI on master. Should be fixed now. Could you please rebase?

fommil:
done

feuerbach:
Thanks, merged and released as tasty-1.2.1.

PR #236, opened 2019-01-09
Add tasty-wai to provider list

mankyKitty:
I've written a provider for tasty that can be used to test the API in a `wai` application: https://hackage.haskell.org/package/tasty-wai

Wanted to add it to the list of providers, apologies if there is a more formal process that I've missed!

I also changed every `http` I could see to `https`. I tested the links and nothing appears to be broken.


feuerbach:
Sweet, thanks!

Issue #48, opened 2014-01-29
Inter-test dependencies


feuerbach:
@jstolarek writes:

> I want to run test A first and if it succeeds I want to run test B (if A fails B is skipped).

@joeyh has also requested this feature in #47.

The main problem here is a syntax/API for declaring dependencies. It would be relatively easy for hspec-style monadic test suites, but I don't see a good way to do it using the current list-based interface.

(This could be a reason to move to the monadic interface. But it's a big change, and I'd like to explore other options first.)

mightybyte:
I think I agree with @ocharles that the monadic interface feels like the way to go here.  It seems like we need the most powerful abstraction possible at the lowest level, otherwise we keep end up running into these situations where as [Ed says](https://www.youtube.com/watch?v=yFXzuCFeRGM&feature=youtu.be&t=1h34m55s), "we can't say all the things".

quchen:
Bump because I’ve been missing this feature a lot for a long time now!

```haskell
onSuccessRun :: TestTree -> TestTree -> TestTree
onSuccessRun precondition moreTests = ...
```

feuerbach:
@quchen as I said above,

> I don't have resources to work on this atm (and in the foreseeable future), but you can totally try it yourself. I'm happy to review your early designs/prototypes.

The API you propose looks fine to me.

quchen:
Oh, I missed that. But it’s good to know you’re open to such a feature, so contributors can be sure it is merged once they get it done. :-)

feuerbach:
I've been slowly working on this feature, and now it's pretty much finished (except documentation) in the `dependencies` branch.

With it, you can apply `after AllFinish "pattern"` or `after AllSucceed "pattern"` to a test tree, and it will make all the tests inside the tree execute only after all tests whose name match the pattern have finished (respectively, succeeded).

Here's a minimal example:

``` haskell
{-# LANGUAGE NumDecimals #-}
import Test.Tasty
import Test.Tasty.HUnit
import Control.Concurrent

main = defaultMain $ testGroup "Tests" $
  [ after AllSucceed "Three" $ testCase "One" $ threadDelay 1e6
  , testCase "Two" $ threadDelay 1e6
  , testCase "Three" $ threadDelay 1e6 >> assertFailure "fail"
  ]
```

Let me know what you think.

RyanGlScott:
[...]

PR #249, opened 2019-05-05
Wait for tests to release their resources after Ctrl-C

gelisam:
`tasty` already has `withResource` for acquiring a resource which is shared by multiple tests, but within a single test, `bracket` and `finally` should still work as expected, right? Well, they don't :(

In the following simple example, if I hit Ctrl-C between `sleeping...` and `slept`, `releasing...` gets printed, but then the program terminates immediately without printing `released`. If I remove the `defaultMain $ testCase "acquire, sleep, release."` part and hit Ctrl-C at the same point, I get the expected behaviour: `releasing...` gets printed, the program hangs for 2 seconds, and then prints `released`.

    import Test.Tasty
    import Test.Tasty.HUnit
    
    import Control.Concurrent
    import Control.Exception
    
    main :: IO ()
    main = defaultMain $ testCase "acquire, sleep, release." $ do
      finally (putStrLn "sleeping..." >> threadDelay 2000000 >> putStrLn "slept")
              (putStrLn "releasing..." >> threadDelay 2000000 >> putStrLn "released")

This PR includes a change which fixes the problem ~in this case, but even though I can see that the problem is fixed when I use Ctrl-C, I am having some difficulties writing a test case. I spent the day on this and I still couldn't figure it out, so I am submitting what I have so far as a PR in the hope that you can help me finish it.~

### DIFF
--- a/core-tests/resource-release-test.sh
+++ b/core-tests/resource-release-test.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+
+set -ux
+
+# This assumes that the compiled resource-release-test executable
+# is in the PATH
+
+# create a folder for the duration of the test, and make sure it starts empty
+DIR=`dirname $0`/resource-release-test-files
+rm -rf "$DIR"
+mkdir "$DIR"
+function finish {
+  rm -rf "$DIR"
+}
+trap finish EXIT
+
+# start resource-release-test and wait until it's ready
+resource-release-test "$DIR" &
+PID="$!"
+while [ ! -f "$DIR/test-has-started" ]; do sleep 1; done
+
+# kill resource-release-test and wait for it to release its resources.
+# setsid runs 'kill' in a new process group, otherwise the signal doesn't get
+# delivered.
+kill -s INT "$PID"
+wait "$PID"
+
+# check that the resources were correctly released
+[ -f "$DIR/resources-are-released" ] || exit 1
--- a/core-tests/resource-release-test.sh
+++ b/core-tests/resource-release-test.sh
@@ -6,10 +6,10 @@ set -ux
 # is in the PATH
 
 # create a folder for the duration of the test, and make sure it starts empty
-DIR=`dirname $0`/resource-release-test-files
+DIR="$(dirname "$0")/resource-release-test-files"
 rm -rf "$DIR"
 mkdir "$DIR"
-function finish {
+finish() {
   rm -rf "$DIR"
 }
 trap finish EXIT
--- a/core-tests/resource-release-test.sh
+++ b/core-tests/resource-release-test.sh
@@ -20,8 +20,6 @@ PID="$!"
 while [ ! -f "$DIR/test-has-started" ]; do sleep 1; done
 
 # kill resource-release-test and wait for it to release its resources.
-# setsid runs 'kill' in a new process group, otherwise the signal doesn't get
-# delivered.
 kill -s INT "$PID"
 wait "$PID"
 
--- a/core/Test/Tasty/Parallel.hs
+++ b/core/Test/Tasty/Parallel.hs
@@ -55,7 +55,11 @@ runInParallel nthreads actions = do
 
   pids <- replicateM nthreads (async $ work actionsVar)
 
-  return $ mapM_ cancel pids
+  return $ do
+    -- tell worker threads there is no more work after their current task
+    _ <- atomically $ swapTMVar actionsVar []
+    -- cancel all the current tasks, waiting for workers to clean up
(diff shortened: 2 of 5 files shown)


core/Test/Tasty/Parallel.hs
```diff
@@ -52,9 +53,9 @@ runInParallel nthreads actions = do
 
   actionsVar <- atomically $ newTMVar actions
 
-  pids <- replicateM nthreads (forkIO $ work actionsVar)
+  pids <- replicateM nthreads (async $ work actionsVar)
 
-  return $ mapM_ killThread pids
+  return $ mapM_ cancel pids
```
gelisam: I believe the reason the `bracket` block doesn't release its resources is because the main thread terminates too early. So my suggested fix is to switch to `async` (which is already a dependency) so we can benefit from `cancel`'s guarantee that it only returns once the thread has completed, that is, once all the resources are released.

core-tests/resource-release-test.sh
```diff
@@ -0,0 +1,27 @@
...
+}
+trap finish EXIT
+
+# start resource-release-test and wait until it's ready
+resource-release-test "$DIR" &
+PID="$!"
+while [ ! -f "$DIR/test-has-started" ]; do sleep 1; done
+
+# kill resource-release-test and wait for it to release its resources.
+kill -s INT "$PID"
```
gelisam: I am trying to reproduce the "hit Ctrl-C while the test is sleeping" behaviour by sending a SIGINT signal when I know, via other means, that this is where the test's execution is stuck. Tracing indicates that this successfully causes the `finally` block to release its resources, but for reasons I don't understand, `defaultMain` doesn't terminate once the exception finishes propagating up the stack. So this test never terminates.

core/Test/Tasty/Parallel.hs
```diff
@@ -52,9 +53,9 @@ runInParallel nthreads actions = do
...
 
 work :: TMVar [Action] -> IO ()
 work actionsVar = go
```
gelisam: I tried replacing the `actionRun this :: IO ()` below with the contents of my test, and that fixed the problem, `defaultMain` terminates again. Which means that `actionRun this` is not simply running one of the tests, as I expected, it is instead doing something more complicated. I am still reading the code and #48 in order to figure out what it is that `actionRun this` is doing instead.

gelisam:
I figured it out! When I used `Async.cancel` to stop the worker threads, it didn't stop the worker threads, it only caused an exception to be raised during a test, and so the workers marked that test as failed but moved on to the next task. There was no next task, but the first worker to find a full `TMVar` with an empty list of tasks left the `TMVar` empty, thus preventing the other threads from using `takeTMVar` to also discover that the action list is empty.

I have fixed both problems: upon encountering Ctrl-C, I now fill the `TMVar` with an empty list of actions so that the worker threads don't continue working after marking their current test as failed, and the first worker to find a full `TMVar` with an empty list of tasks fills the `TMVar` back with that empty list of tasks. `defaultMain` no longer hangs, my test passes, and tests are now given the time to release their resources when the test suite aborts due to a Ctrl-C.

core/Test/Tasty/Parallel.hs
gelisam: I now understand that `actionRun this` does run one of the tests, but it runs it inside a `try`, and so `Async.cancel` does not cause the worker thread to terminate.

core-tests/resource-release-test.sh
gelisam: I now understand that `defaultMain` wasn't terminating because the worker threads were not terminating, see my top-level comment.

core/Test/Tasty/Parallel.hs
```diff
@@ -63,8 +68,9 @@ work actionsVar = go
       join . atomically $ do
         mb_ready <- findBool =<< takeTMVar actionsVar
         case mb_ready of
-          Nothing ->
+          Nothing -> do
             -- nothing left to do; return
+            putTMVar actionsVar []
```
gelisam: the `putTMVar` allows the other worker threads to return as well.

gelisam:
bump! any comment on this PR?

feuerbach:
Sorry, I just haven't had the time to look at this. Might happen later this week, but no promises.

gelisam:
no rush, just poking once in a while to make sure it doesn't fall off the radar :)

feuerbach:
Thanks @gelisam, I finally got around to this. Excellent analysis and a neat fix.

feuerbach:
Released as tasty-1.2.3.

PR #252, opened 2019-05-17
Expose computeStatistics from ConsoleReporter

### LINKED ISSUE #251: Expose computeStatistics from Ingredients.ConsoleReporter
This is similar to #183, but I wish to expose more of the internals. 

In particular, I want to expose `computeStatistics`. There is `Statistics` exposed, as well as `printStatistics`, but not `computeStatistics`.

This is for a custom ingredient that shortens test case outputs when printing, but serializes the full test case to disk ([here is a link to the ingredient](https://github.com/tweag/asterius/pull/125/files#diff-e5a74906999b7f2b20abb90223c90eb0R201))

Would a PR that exposes this function be acceptable?

Thanks a lot!

bollu:
Closes #251 

Adds `computeStatistics` to the export list of the module.

### DIFF
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -15,6 +15,7 @@ module Test.Tasty.Ingredients.ConsoleReporter
   , useColor
   -- ** Test failure statistics
   , Statistics(..)
+  , computeStatistics
   , printStatistics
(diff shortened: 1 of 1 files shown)


feuerbach:
Do you need a release for this?

bollu:
It would be nice to have one, but I understand if that's a pain :)

feuerbach:
This is now released as tasty-1.2.3.

PR #265, opened 2019-10-22
Add optional provider controlled detailed result reports

mbj:
### Motivation

I'm in the need for a golden test framework that renders a "pretty" (as with colors) unified diff output on expectation mismatches. And I made a test provider for this use case.

But: I think there is no way for the tasty console reporter to render such a report. Hence I took my stab to extend the tasty core with such an API.

### On the commits

* All but the last commits are cleanups "I like", I'm fine to drop all of them in need. Just let me know.
* The last commit is the feature addition.

### Notes on the last commit

* The `printResultDetails` function has a rather complex signature, this is because I wanted the provider I write to not re-do any console details discovery.
* I expect no changes to existing providers.
* I expect no runtime overhead for existing providers.
* I "currently" do not have an idea to make a less invasive API, and I'm very happy to get pointed to improvements.
* I also avoided to pass the current indentation to the detailed report, as I do not need it in my case.

I'm very happy to receive input on this PR and address any flags that come up.

### DIFF
--- a/core/Test/Tasty/ConsoleFormat.hs
+++ b/core/Test/Tasty/ConsoleFormat.hs
@@ -1,0 +1,24 @@
+module Test.Tasty.ConsoleFormat
+  ( ConsoleFormat(..)
+  , failFormat
+  , infoFailFormat
+  , infoOkFormat
+  , okFormat
+  , skippedFormat
+  )
+where
+
+import System.Console.ANSI
+
+data ConsoleFormat = ConsoleFormat
+  { consoleIntensity :: ConsoleIntensity
+  , colorIntensity   :: ColorIntensity
+  , color            :: Color
+  }
+
+failFormat, infoFailFormat, infoOkFormat, okFormat, skippedFormat :: ConsoleFormat
+failFormat     = ConsoleFormat BoldIntensity   Vivid Red
+infoFailFormat = ConsoleFormat NormalIntensity Dull  Red
+infoOkFormat   = ConsoleFormat NormalIntensity Dull  White
+okFormat       = ConsoleFormat NormalIntensity Dull  Green
+skippedFormat  = ConsoleFormat NormalIntensity Dull  Magenta
--- a/core/Test/Tasty/ConsoleFormat.hs
+++ b/core/Test/Tasty/ConsoleFormat.hs
@@ -1,5 +1,11 @@
+-- | Console output format
+--
+-- These console formats tasty uses when reporing.
+-- They are exported so custom providers with custom report printing
+-- can re-use the tasty output formats
 module Test.Tasty.ConsoleFormat
   ( ConsoleFormat(..)
+  , ConsoleFormatPrinter
   , failFormat
   , infoFailFormat
   , infoOkFormat
@@ -10,15 +16,35 @@ where
 
 import System.Console.ANSI
 
+-- | Console output format
 data ConsoleFormat = ConsoleFormat
   { consoleIntensity :: ConsoleIntensity
   , colorIntensity   :: ColorIntensity
   , color            :: Color
   }
 
-failFormat, infoFailFormat, infoOkFormat, okFormat, skippedFormat :: ConsoleFormat
-failFormat     = ConsoleFormat BoldIntensity   Vivid Red
+-- | Type of console format printer functions
+type ConsoleFormatPrinter
+  =  ConsoleFormat -- ^ selected console format
+  -> IO ()         -- ^ action to be executed with active console format
+  -> IO ()
+
+-- | Format used to display failures
+failFormat :: ConsoleFormat
+failFormat = ConsoleFormat BoldIntensity   Vivid Red
+
+-- | Format used to display additional information on failures
+infoFailFormat :: ConsoleFormat
 infoFailFormat = ConsoleFormat NormalIntensity Dull  Red
-infoOkFormat   = ConsoleFormat NormalIntensity Dull  White
-okFormat       = ConsoleFormat NormalIntensity Dull  Green
-skippedFormat  = ConsoleFormat NormalIntensity Dull  Magenta
+
+-- | Format used to display sucesses
+okFormat :: ConsoleFormat
+okFormat = ConsoleFormat NormalIntensity Dull  Green
+
+-- | Format used to display additional information on sucesses
+infoOkFormat :: ConsoleFormat
+infoOkFormat = ConsoleFormat NormalIntensity Dull  White
+
+-- | Format used to display skipped tests
+skippedFormat :: ConsoleFormat
+skippedFormat = ConsoleFormat NormalIntensity Dull  Magenta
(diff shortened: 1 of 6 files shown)


mbj:
I see now on how the CI does build on older GHC. 

And I suspect not all commits in here will pass with these. I'll wait for the CI to finish and than rewrite the commit sequence to pass with the build matrix on Travis.

mbj:
As support for older GHC was dropped I rebased it. Lets see if it gets green.

mbj:
@feuerbach As the build is now green: Can I interest you to review of this PR?

What do I miss to get a feature such as I'm requesting here merged?

core/Test/Tasty/Core.hs
```diff
@@ -65,8 +68,22 @@ data Result = Result
...
     -- ^ How long it took to run the test, in seconds.
+  , printResultDetails :: ResultDetailsPrinter
+    -- ^ Optional detailed result printing action.
+    --
+    -- 'printResultDetails' is used by providers that need to have fine grained
+    -- control about the printed results. Especially when it comes to colorization.
   }
-  deriving Show
+
+instance Show Result where
```
mbj: This instance exists as the `ResultDetailsPrinter` itself has no `Show` instance.

An alterative would be to wrap it into a type with explicit "dummy" show instance. I'm not sure which one you'd prefer.

core/Test/Tasty/ConsoleFormat.hs
```diff
@@ -0,0 +1,24 @@
+module Test.Tasty.ConsoleFormat
```
mbj: This module was added so that providers can render in the tasty console format without having to re-do some of its logic.

core/Test/Tasty/Ingredients/ConsoleReporter.hs
```diff
@@ -623,28 +625,33 @@ stringWidth = length
 
 -- (Potentially) colorful output
 ok, fail, skipped, infoOk, infoFail :: (?colors :: Bool) => String -> IO ()
-fail     = output BoldIntensity   Vivid Red
-ok       = output NormalIntensity Dull  Green
-skipped  = output NormalIntensity Dull  Magenta
-infoOk   = output NormalIntensity Dull  White
-infoFail = output NormalIntensity Dull  Red
+fail     = output failFormat
```
mbj: These changes where done to deduplicate with the added `ConsoleFormat` module.

feuerbach:
> @feuerbach As the build is now green: Can I interest you to review of this PR?
> 
> What do I miss to get a feature such as I'm requesting here merged?

Sorry @mbj, you did nothing wrong, I just haven't had the bandwidth to look into this. I'll add this to my todo list, but I can't promise anything.

mbj:
> I just haven't had the bandwidth to look into this. I'll add this to my todo list, but I can't promise anything.

No worries! I know that situation. If there is anything *I can do* to help you let me know. I'm a maintainer of some semi-popular OSS by myself and know that maintainer time is super limited. So if I can do something to *help you*: Let me know.

mbj:
I just updated this PR with a newtype around the detailed console printer to remove the ugly manual show instance of results.

Also I could cut down the duplication a bit when defaulting to a noop details printer.

mbj:
@feuerbach I'm very happy with your feedback. I'll address them.

One suggestion to reduce the review load on you, which you try to address via asking me to add commits instead to re-write the original commit.

What I think it may help if I where to extract the `ConsoleReporter` to a separate PR (putting it at the correct location you indicated).

This change can likely be merged before my main change and would reduce the review load?

mbj:
Same for the whitespace and vim fix commits. I could PR them early in a cleanup PR. Reducing the inventory of this PR.

feuerbach:
Thanks — I do the review & merging locally anyway, so having a single PR is best for me.

mbj:
> 1. Please add haddocks to the new types and top-level functions that you've added. Also, the new `ConsoleFormat` should have some module-level haddock briefly explaining to the user the purpose of this module and how it can be used.

done. The last commit where I added the haddocs also had me to add a bunch of type aliases to make the "callback" function arguments easier to document. As haddock does AFAIK not like to document the `a` / `b` in functions like: `(a -> b) -> c`. Overall the code should be nicer with that refactorings.

> 2. I pushed a commit to master to remove the foldtext configuration (which also causes issues for me). I left the fdm=marker configuration, as it's doesn't really belong to the user's global config: it only makes sense to enable for a file that contains vim's folding markers (which no other file in tasty does).

Thanks, I've dropped my version of the commit.

> 3\. Should `ResultDetailsPrinter` additionally accept the current indentation level, so that its output can be indented similarly to the rest of tasty's output?

I added the current indentation level as `Int` (just like the main reporter tracks it) to the function wrapped inside the `ResultDetailsPrinter`. This can allow provider authors to also indent their output. I'll not use it in my provider.

> 4\. Since `Test.Tasty.ConsoleFormat` is to be used primarily by providers (as far as I understand?), I think it belongs under the `Providers` hierarchy, i.e.  `Test.Tasty.Providers.ConsoleFormat`.

The `Test.Tasty.ConsoleFormat` module defines re-usable types also used by tasty core reporter. 

After performing the move I found that `Test.Tasty.Core` would import `Test.Tasty.Provider.ConsoleFormat` which does not really strike me as the correct solution, why should the "core" import the provider interface? Hence I left the module at its original place.

> 5. I think a provider author may be confused by the relationship between resultDescription and resultDetailsPrinter: can you use both, how do they interact, when should you use one vs the other vs both? Maybe you could provide some guidance, e.g. in the top-level haddock for the Result type.

I've added detailed haddocs for both fields also referencing each other.

>  (in new commit(s), so that it's easier for me to look just at the new changes):

I did so. But not note: I rebased on top of master to drop the comment commit as discussed under 2). Also normally I'd never PR a commit sequence like this. Simply because my earlier commits introduce problems my later commits solve (missing haddocs). This should not happen inside a single PR its not "hygenic" Hence I recommend you squash before merging. I see the problem more on githubs side, they should allow to diff the state since you last saw it, but not force me to make "unhygenic" commits. For now I did like you asked and make "dirty ones".

mbj:
@feuerbach Friendly ping. No pressure intend. Take your time. Cheers.

feuerbach:
Yes, I remember about this. Was going to take a look this weekend (but no promises).

mbj:
> Yes, I remember about this.

Thanks to confirm.

> Was going to take a look this weekend (but no promises).

This is OSS (imagine "this is sparta meme here"). I do not even would want promises ;)

feuerbach:
I merged the PR and added a few minor changes. Once you update your provider and confirm that everything works as expected, I can make a release if you want.

mbj:
@feuerbach Thanks! I'll confirm my provider this evening. Will check back here to ask for a release.

After your release and some polishing and peer review I'll also release my provider to the OSS.

Its basically like `tasty-golden` but with multi line diff prints and slightly different ergonomics / API. I intend to call it `tasty-mgolden`.

> added a few minor changes.

I like these, should sorry to not anticipate them.

mbj:
@feuerbach My provider works with your master HEAD commit. Feel free to release any time. Much apprechiated.

feuerbach:
Released as tasty-1.3.1

PR #270, opened 2020-04-29
Add showDefaultValue to IsOption for default argument rendering in --help

### LINKED ISSUE #269: Make displaying defaultValue in --help output more idiomatic
`tasty` has an `IsOption` class that abstracts the details of how command-line options are handled. One thing is currently does not do is handle how [`defaultValue`](https://hackage.haskell.org/package/tasty-1.2.3/docs/Test-Tasty-Options.html#v:defaultValue)s are displayed in the `--help` output. As a result, this is usually tacked on to the end of `optionHelp` manually. Here are three examples from `tasty` alone:

https://github.com/feuerbach/tasty/blob/79132dd4f270e2bb6804c7d6dcdf8d5d07d204a4/core/Test/Tasty/Options/Core.hs#L38
https://github.com/feuerbach/tasty/blob/79132dd4f270e2bb6804c7d6dcdf8d5d07d204a4/core/Test/Tasty/Ingredients/ConsoleReporter.hs#L461
https://github.com/feuerbach/tasty/blob/79132dd4f270e2bb6804c7d6dcdf8d5d07d204a4/core/Test/Tasty/Ingredients/ConsoleReporter.hs#L485-L488

It would be nice to specify this information declaratively. In fact, `optparse-applicative` has a [`showDefaultWith`](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative.html#v:showDefaultWith) combinator for this very purpose. Alas, we can't just use that: `showDefaultWith` only works when used in conjunction with the [`value`](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative.html#v:value) combinator, and `value` [does not play nicely with `tasty`'s environment variable detection](https://github.com/feuerbach/tasty-golden/issues/37).

Perhaps we can accomplish the same end result without `value`/`showDefaultWith`, however. Here is one proposal for how to go about this:

1. Add a new method to `IsOption`:

   ```hs
   class IsOption v where
     ...
     optionHelpDefault :: v -> Maybe String
     optionHelpDefault _ = Nothing
   ```

   If `optionHelpDefault` returns `Just m`, the `--help` output will display `(default: <m>)` after the `optionHelp` text. If it returns `Nothing`, only the `optionHelp` text will be displayed.
2. In order to make `optparse-applicative` actually display `optionHelpDefault`, we'll need to change [`propShowDefault`](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative-Types.html#v:propShowDefault) accordingly. It's tempting to think that [`optionMod`](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative-Builder-Internal.html#v:optionMod) would suffice to accomplish this, but in practice, that gets overwritten based on the values in [`DefaultProp`](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative-Builder-Internal.html#t:DefaultProp). A possible alternative to modify the final `Parser` after it is built.
3. Finally, to ensure that we don't accidentally shoot ourselves in the foot, we should outlaw the use of `value` in `optionCLParser`. This, too, might be possible by inspecting the final `Parser` after it is built.

   We might consider having a similar check for any uses of `showDefaultWith`, although if we implement this correctly, the effects of `showDefaultWith` will simply be ignored altogether.

Does this sound like a reasonable design, @feuerbach?

RyanGlScott:
This introduces a new `showDefaultValue` method to `IsOption` for the purpose of making it possible to display `defaultValue`s in `--help` output declaratively. For various reasons, `optparse-applicative`'s own `showDefaultValue` combinator is unsuitable for this purpose (see the discussion in #269), so we do the next best thing by rolling our own way to do it.

This approach isn't perfect, as there are various ways you can shoot yourself in the foot if you construct a `Parser` recklessly. For this reason, this patch also adds various validity checks that issue warnings (or errors) if you do silly things. See `Test.Tasty.Options.modValidityChecks` and `Test.Tasy.CmdLine.warnMultipleOptPs`.

Fixes #269.


core/Test/Tasty/Options.hs
```diff
@@ -143,13 +152,53 @@ mkOptionCLParser mod =
...
+  = error $ unlines
+      [ "ERROR" ++ prov
+      , "Using Options.Applicative.value (or derived functions) in"
+      , "optionCLParser is prohibited, as it interferes with tasty's ability"
+      , "to read environment variable options properly. Moreover, using the"
+      , "value function is unnecessary, as its functionality is subsumed by"
+      , "defaultValue method of IsOption."
+      ]
+  | isJust mbShowDef
+  = trace (unlines
```
RyanGlScott, line 190: Using `trace` to display warnings is a bit grimy. But since `modValidityChecks` needs to be run from `mkFlagCLParser`/`mkOptionCLParser` (which are pure functions), I'm not sure how to do better without changing their types to involve `IO`.

RyanGlScott:
I've tried out the behavior of the errors/warnings on a small example, but it's not clear if I can add this example to the test suite somehow.

core/Test/Tasty/Options.hs
feuerbach: You can use e.g. a Writer monad to collect the errors, and then print the errors once we reach IO.

RyanGlScott: I'm not sure how that would work in the context of `mkFlagCLParser`/`mkOptionCLParser`, as these exist to implement `optionCLParser`. Unless you are suggesting to change the type of `optionCLParser` to include `Writer`?

feuerbach: It doesn't have to be part of `optionCLParser`. It should be a separate post-processing function applied on top of `optionCLParser` inside the `optionParser` function (whose type will change to also return a list of warnings).

feuerbach: In other words, we shouldn't rely on the user using mkFlagCLParser or otherwise including modValidityChecks. Instead, we take whatever parser the user provided, and validate and amend it w.r.t. the default values.

RyanGlScott: Unfortunately, it is not possible to perform the validity checks in `modValidityChecks` on a `Parser`. This is because the internal structure of a `Parser a` does not record whether the `a` comes from a command-line argument or a default `value` (see [`mkParser`](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/src/Options.Applicative.Builder.Internal.html#mkParser)). Moreover, a `Parser` only records the presence of `showDefaultWith` if it is used in combination with `value`. If `showDefaultWith` is used by itself, it will be ignored in the internal structure of the `Parser`'s `OptProperties` (see [`mkProps`](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/src/Options.Applicative.Builder.Internal.html#mkProps)).

The only way I know how to accurately gauge this information is to look at a `Mod` before it is passed into `mkParser` and loses its internal structure.

feuerbach: Ah I see now.

In that case I suggest we test for it "functionally": evaluate the Parser applicative on an empty input and see if it succeeds. In fact, it seems like `evalParser` from `Options.Applicative.Common` does exactly that.

RyanGlScott: Wow, I can't believe I overlooked that! That solves the problem of detecting default values quite nicely. Sadly, I still can't figure out a clean way to detect the presence of `showDefaultWith`, but I think I'll just give up trying to warn about that. (I've added a disclaimer to the Haddocks for `optionCLParser` about this.)

RyanGlScott:
I've pushed a major change that removes all uses of `Debug.Trace.trace` (yay!) and shunts all the work of reporting warnings/errors to `parseOptions`, which lives in `IO`. This involved changing the return types of `optionParser` and `suiteOptionParser`, so this is a breaking API change. Let me know if this is to your liking.

feuerbach:
Thanks Ryan — looks good for the most part, but as I said in https://github.com/feuerbach/tasty/issues/269#issuecomment-621037141, I'd really prefer these to be warnings just in case someone's code is already doing these things. Could you convert everything to warnings?

RyanGlScott:
Whoops, I overlooked the "in all these cases"  part of that comment. I've replaced `OptionMessages` with a simple `[String]` of warnings.

feuerbach:
Thanks Ryan! I released this as tasty-1.3 and made a corresponding change to tasty-golden.

RyanGlScott:
Wonderful! Thanks for the detailed review feedback.

PR #279, opened 2020-10-20
Allow `TestPattern` to be modified inside a `TestTree`

dcastro:
I ran into a scenario where I needed to modify the `TestPattern` option programmatically.

It seems Tasty does not support this, because `foldTestTree` reads the pattern once at the start, before traversing the tree, and then never reads it again.

This PR changes `foldTestTree` to read the pattern option each time it goes down a level.


feuerbach:
Nice, thanks!

PR #287, opened 2020-12-10
Restore compatibility back to GHC 7.0

Bodigrim:
This restores compatibility of `tasty`, `tasty-hunit` and `tasty-quickcheck` with old GHCs, back to GHC 7.0. 

My applications do not require `tasty-smallcheck`, so I left it as is. If desired, it could be made compatible with GHC 7.0 as well, see https://github.com/Bodigrim/tasty/commit/003b2a39888ff7f4d8a0d550feff1e7491bfa184, but it involves pretty significant rewrite.

I also kept Travis config as is, but if you wish to test backward compatibility regularly, I can submit additional jobs. Here is a witness that the libraries actually build on GHC 7.0: https://travis-ci.com/github/Bodigrim/tasty/builds/207660317


feuerbach:
These are released now.

Bodigrim:
Thanks!

PR #293, opened 2021-02-05
Deduplicate command-line options

### LINKED ISSUE #292: Duplicate command-line options
if there is more than one `TestReporter` then `--help` lists `-j,--num-threads` several times:

https://github.com/feuerbach/tasty/blob/23d856cdce96dd71b825490da30b4338a4f0a582/core/Test/Tasty/Ingredients.hs#L108-L115

Similarly, if the same command-line option is shared between several providers / ingredients, it ends up reported multiple times as well:

https://github.com/feuerbach/tasty/blob/23d856cdce96dd71b825490da30b4338a4f0a582/core/Test/Tasty/Ingredients.hs#L119-L123

Cf. https://github.com/Bodigrim/tasty-bench/issues/3

I'm happy to raise a PR fixing this, but unsure about approach. Should we try to preserve the order of options? Or throwing everything in a `Set` and getting back alphabetically sorted will do?

Bodigrim:
Closes #292.

### DIFF
--- a/core/Test/Tasty/Ingredients.hs
+++ b/core/Test/Tasty/Ingredients.hs
@@ -112,12 +112,12 @@ ingredientOptions (TestManager opts _) = opts
 
 -- | Like 'ingredientOption', but folds over multiple ingredients.
 ingredientsOptions :: [Ingredient] -> [OptionDescription]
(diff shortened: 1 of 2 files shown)


feuerbach:
Thanks, released as tasty-1.4.1.

PR #300, opened 2021-04-20
Fix dead link

kodaj:
jstolarek.com no longer exists, but a Google search for "Code testing in Haskell revisited (with Tasty)" gives an alternative link.


UnkindPartition:
Thanks!

PR #303, opened 2021-06-28
Patches for GHC 9.4

Bodigrim:
GHC 9.4 is not coming any time soon, but patches are tiny and fully backward compatible (see https://travis-ci.com/github/Bodigrim/tasty/builds/231156026), so I guess it should not hurt to merge them early. Sometimes it is important to run tests on bleeding-edge GHC.

Cf. https://gitlab.haskell.org/ghc/head.hackage/-/blob/master/patches/tasty-1.4.1.patch
CC @RyanGlScott


UnkindPartition:
Thanks!

PR #304, opened 2021-06-28
Suggest pattern to rerun an individual failing test

### LINKED ISSUE #271: Suggest pattern to rerun an individual failing test
I would like `tasty` to suggest me an appropriate `--pattern PATTERN` to rerun an individual failing test, similar to [hspec](http://hspec.github.io/match.html). Is there an ingredient available? If not, could it be implemented as a part of `ConsoleReporter` or should I go for a standalone ingredient in a separate package?

Bodigrim:
Closes #271. CI build is available at https://travis-ci.com/github/Bodigrim/tasty/builds/231156314

Unfortunately, it seems difficult to implement an idea from https://github.com/UnkindPartition/tasty/issues/271#issuecomment-782822055. The reason is that we do not necessarily have an access to a full set of test names: our initial run could have been constrained by an independent `--pattern`.

### DIFF
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -35,8 +35,10 @@ import Test.Tasty.Core
 import Test.Tasty.Providers.ConsoleFormat
 import Test.Tasty.Run
 import Test.Tasty.Ingredients
+import Test.Tasty.Ingredients.ListTests
 import Test.Tasty.Options
 import Test.Tasty.Options.Core
+import Test.Tasty.Patterns
 import Test.Tasty.Patterns.Printer
 import Test.Tasty.Patterns.Types
 import Test.Tasty.Runners.Reducers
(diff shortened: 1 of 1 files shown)


UnkindPartition:
Couldn't we query the current pattern and `&&` it with the inferred minimal name pattern?

It looks like we don't have a pretty printer for awk expressions; feel free to add one, or I might find some time to write one if you prefer.

Bodigrim:
@UnkindPartition good point, seems working. I've implemented a very naive algorithm for a minimal pattern, not sure whether something more sophisticated is warranted.

Bodigrim:
Ideally I'd like to avoid single quotes around generated patterns when possible (e. g., no spaces or brackets inside), but not sure how to check for such possibility in a terminal-agnostic way.

Bodigrim:
Pushed an updated version, improving performance (do not compute `length` and do not `filter` strings, which have already failed). @UnkindPartition how does it look?

UnkindPartition:
This looks great, thanks!

Bodigrim:
@UnkindPartition I'd appreciate a release including this patch, when you have time.

Bodigrim:
Oops, there is an issue when a test name contains single quotes. I'll prepare a fix.

UnkindPartition:
> Oops, there is an issue when a test name contains single quotes. I'll prepare a fix.

Exactly, I was going to fix that before the release, and also merge some of the other outstanding PRs. The fix should be something like replacing all `'` with `'\''`.

UnkindPartition:
@Bodigrim released as tasty-1.4.2!

PR #307, opened 2021-08-09
Added quickcheck-chatty option

coot:
There are some problems with it.  When running multiple tests from a test suite
the console output becomes messy (is there a way to fix this?), nevertheless
I find this option useful when running a specific test with a large value of
`--quickcheck-tests`.


UnkindPartition:
> is there a way to fix this?

Not really — see <https://github.com/UnkindPartition/tasty#faq>.

I wonder if https://github.com/UnkindPartition/tasty/pull/245 does what you need. Perhaps you could rebase/review/revive that PR instead?

coot:
Subsumed by #311.

Issue #309, opened 2021-08-30
Communicate SIGINT to the parent shell


UnkindPartition:
See <https://www.cons.org/cracauer/sigint.html>

santiweight:
Heya - I'm havingt his issue quite often and it's really annoying - is there somewhere you could point me to in the code to fix it?

UnkindPartition:
Yeah, I had a branch with a fix, but something funky was going on when I ran an automated test for it...

I'll try to dig up and push that branch next time I have access to the computer (https://github.com/UnkindPartition/UnkindPartition/blob/main/README.md) and maybe someone can finish it.

PR #331, opened 2022-05-06
Communicate SIGINT to the parent shell

UnkindPartition:
This is the current work I have for #309. Please review and test it well before merging; I don't remember (and can't test right now) if it actually worked.


santiweight:
Thank you for this! I will try to test this out on Monday at work. If I don't respond in a couple days and you still want review, please ping :)

VictorCMiraldo:
I'd love to write a test for this, but my attempts behave the same at `master` and on this branch. I've tried:
- Write a testGroup that launches processes, then manually send `SIGINT`: I see the entire process killed and get my shell back in both branches
- Write a testGroup with one test that sends `SIGINT` to `$$`, only that test fails on both branches

PR #340, opened 2022-06-20
Print dependency cycles in error message

martijnbastiaan:
While Tasty detects cycles and prints an error message, it does not print the cycles it found. This PR makes sure it does. I've tested it on another code base with cycles in it, the output looks like:

```
clash-testsuite: Test dependencies form a loop:

..tests.shouldwork.Issues.T2046.VHDL.Vivado => ..tests.shouldwork.Issues.T2046.VHDL.Vivado

..tests.shouldwork.Issues.T2046.SystemVerilog.Vivado => ..tests.shouldwork.Issues.T2046.SystemVerilog.Vivado

..tests.shouldwork.BlackBox.T2117.VHDL.Vivado => ..tests.shouldwork.BlackBox.T2117.VHDL.Vivado

..tests.shouldfail.Verification.NonTemporalPSL.VHDL.Vivado => ..tests.shouldfail.Verification.NonTemporalPSL.VHDL.Vivado
```

Note the double dot is an artifact of this particular test tree.

### DIFF
--- a/core/Test/Tasty/Run.hs
+++ b/core/Test/Tasty/Run.hs
@@ -1,6 +1,7 @@
 -- | Running tests
 {-# LANGUAGE ScopedTypeVariables, ExistentialQuantification, RankNTypes,
-             FlexibleContexts, BangPatterns, CPP, DeriveDataTypeable #-}
+             FlexibleContexts, BangPatterns, CPP, DeriveDataTypeable,
+             LambdaCase #-}
 module Test.Tasty.Run
   ( Status(..)
   , StatusMap
@@ -12,6 +13,7 @@ import qualified Data.IntMap as IntMap
 import qualified Data.Sequence as Seq
 import qualified Data.Foldable as F
 import Data.Maybe
+import Data.List (intercalate)
 import Data.Graph (SCC(..), stronglyConnComp)
 import Data.Typeable
 import Control.Monad (forever, guard, join, liftM)
@@ -219,14 +221,21 @@ type Tr = Traversal
 
 -- | Exceptions related to dependencies between tests.
 data DependencyException
-  = DependencyLoop
-    -- ^ Test dependencies form a loop. In other words, test A cannot start
+  = DependencyLoop [[Path]]
+    -- ^ Test dependencies form cycles. In other words, test A cannot start
     -- until test B finishes, and test B cannot start until test
-    -- A finishes.
+    -- A finishes. Field lists detected cycles.
   deriving (Typeable)
 
 instance Show DependencyException where
-  show DependencyLoop = "Test dependencies form a loop."
+  show (DependencyLoop css) = "Test dependencies have cycles:\n" ++ showCycles css
+    where
+      showCycles = intercalate "\n" . map showCycle
+      showPath = intercalate "." . F.toList
+
+      -- For clarity in the error message, the first element is repeated at the end
+      showCycle []     = "- <empty cycle>"
+      showCycle (x:xs) = "- " ++ intercalate ", " (map showPath (x:xs ++ [x]))
 
 instance Exception DependencyException
 
@@ -252,14 +261,14 @@ createTestActions opts0 tree = do
         opts0 tree
   (tests, fins) <- unwrap (mempty :: Path) (mempty :: Deps) traversal
   let
-    mb_tests :: Maybe [(Action, TVar Status)]
+    mb_tests :: Either [[Path]] [(Action, TVar Status)]
     mb_tests = resolveDeps $ map
       (\(act, testInfo) ->
         (act (Seq.empty, Seq.empty), testInfo))
       tests
   case mb_tests of
-    Just tests' -> return (tests', fins)
-    Nothing -> throwIO DependencyLoop
+    Right tests' -> return (tests', fins)
+    Left cycles -> throwIO (DependencyLoop cycles)
 
   where
     runSingleTest :: IsTest t => OptionSet -> TestName -> t -> Tr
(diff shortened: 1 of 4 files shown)


martijnbastiaan:
Sorry, goofed up. The PR should now pass CI as demonstrated [here](https://github.com/martijnbastiaan/tasty/pull/1).

Bodigrim (approved):
LGTM. Indeed, debugging dependency cycles is quite tricky currently. 

Since `DependencyLoop` has been exposed to users, this is a breaking change, but I doubt that there are many clients affected.

martijnbastiaan:
I've added a changelog entry. @Bodigrim Given that this PR has been open for two weeks and hasn't attracted any comments, I doubt it is controversial. Is there anything I can do to help this get merged?

Bodigrim:
I think this is a valuable addition, which justifies a minor breakage. Indeed I needed such feature in the past. @andreasabel @ocharles @VictorCMiraldo @adamgundry opinions?

VictorCMiraldo:
I never needed this and I didn't know that `tasty` does detect dependency cycles. Makes sense and I can see how this can useful. The only think I'd suggest is shortening the output, if we have a dependency loop `[a,b,c,d]`, why not print names only once? For instance:

```
Cyclic test dependencies:
  - a
  - b
  - c
  - d
```

martijnbastiaan:
I did that to make self-loops non-ambiguous / more obvious to read. In my example I've got four separate dependency cycles, but they're all forming a self-loop. If we'd go with the list as you posted, I think we should do something like:

```
Cyclic test dependencies:

Cycle 1:
  - a
  - b
  - c

Cycle 2:
  - d
```

core/Test/Tasty/Run.hs
```diff
@@ -219,14 +221,21 @@ type Tr = Traversal
...
+  = DependencyLoop [[Path]]
     -- ^ Test dependencies form a loop. In other words, test A cannot start
     -- until test B finishes, and test B cannot start until test
-    -- A finishes.
+    -- A finishes. Field lists detected cycles.
   deriving (Typeable)
 
 instance Show DependencyException where
-  show DependencyLoop = "Test dependencies form a loop."
+  show (DependencyLoop css) = "Test dependencies form a loop:\n\n" ++ showCycles css
```
VictorCMiraldo, line 231: I see! I didn't realize that the message you posted in the PR description had four cycles. Scratch my previous suggestion. :)
I'd still suggest to change the message to something a little more informative. Something like:

```suggestion
  show (DependencyLoop css) = "Test dependencies have cycles:\n\n" ++ showCycles css
```
This would have helped my understanding of the PR description, for instance.

I'd still be in favor of less characters on the message and a more familiar formatting, something like:
```
Test dependencies have cycles:
- a, b, c, a
- x, y, x
- z, z
```

But that's just a personal preference

martijnbastiaan: Makes sense, I'll ping you when it's done.

martijnbastiaan:
@VictorCMiraldo I've implemented your suggestion. Could you re-review and/or start CI?

VictorCMiraldo:
Thank you @martijnbastiaan, I triggered CI.
@Bodigrim, as you mentioned, this is a breaking change. That being said, I agree that it is very unlikely to break for many users and @martijnbastiaan already prepped the changelog with a relevant entry for a future release of tasty. Do you think this still needs anything or should we merge once CI goes green?

VictorCMiraldo (approved):
LGTM, thanks!

core/Test/Tasty/Run.hs
```diff
@@ -219,14 +221,21 @@ type Tr = Traversal
...
+  = DependencyLoop [[Path]]
+    -- ^ Test dependencies form cycles. In other words, test A cannot start
     -- until test B finishes, and test B cannot start until test
-    -- A finishes.
+    -- A finishes. Field lists detected cycles.
   deriving (Typeable)
 
 instance Show DependencyException where
-  show DependencyLoop = "Test dependencies form a loop."
+  show (DependencyLoop css) = "Test dependencies have cycles:" ++ showCycles css
```
Bodigrim, line 231:
```suggestion
  show (DependencyLoop css) = "Test dependencies have cycles:\n" ++ showCycles css
```

Otherwise the output looks like 
```
mod-bench: Test dependencies have cycles:All.Power.Data.Mod, All.Power.Data.Mod
```

martijnbastiaan: Yes, you're right. And `showCycle` should do the `- ` stuff. Currently it looks like:

```
mod-bench: Test dependencies have cycles:All.Power.Data.Mod, All.Power.Data.Mod
- Foo, Bar, Foo
```

core/Test/Tasty/Run.hs
```diff
@@ -219,14 +221,21 @@ type Tr = Traversal
...
+    where
+      showCycles = intercalate "\n- " . map showCycle
```
Bodigrim, line 233: We need `"\n- "` before the first line as well, otherwise the output looks like
```
mod-bench: Test dependencies have cycles:All.Power.Data.Mod, All.Power.Data.Mod
```

martijnbastiaan:
Thanks @Bodigrim, I've added a test for it too. Should be good now!

Bodigrim:
@VictorCMiraldo feel free to merge.

core/Test/Tasty/Run.hs
VictorCMiraldo: Good catch, thanks!

VictorCMiraldo:
Let's merge then! :)

Issue #342, opened 2022-07-13
Add option to set max line length


brandon-leapyear:
Currently, the test output puts the result statuses all at the same column, which matches the longest test name. This causes the tests to wrap when you have just one really long test name:

```
test 1:
            OK
test 2:
            OK
test 5:
            OK
test 10:
            OK
test 100:
            OK
(fromMySpecialType . toMySpecialType) is idempotent except when someOption
Name=False: OK
some other longish named test:
            OK
```

It'd be nice to set a max line length, whereby the "find longest line" calculation only accounts for lines shorter than the given configuration, and longer lines will just wrap as usual. e.g. with the above output and `--max-line-length=10`:
```
test 1:   OK
test 2:   OK
test 5:   OK
test 10:  OK
test 100: OK
(fromMySpecialType . toMySpecialType) is idempotent except when someOption
Name=False: OK
some other longish named test: OK
```

Bodigrim:
There are plenty of things one might wish to configure in `consoleTestReporter`, but it would be overwhelming for casual users to be presented with a screen-long list of display options which they do not care about. 

I suggest we rather export more helpers from `Test.Tasty.Ingredients.ConsoleReporter`, so that other developers could easily implement their own reporter with required capabilities and options.

brandon-leapyear:
I'm sympathetic to that. Alternatively, what do you think about doing something like GHC's `--help` vs `--show-all-options`? Just fantasizing here, it'd be cool if we could do something like what pytest does:

```
$ stack test --ta --help-full

Mmm... tasty test suite

Usage: my-test-suite [OPTIONS] PATTERN

General options:
  -h,--help                Show this help text
  -t,--timeout DURATION    Timeout for individual tests (suffixes: ms,s,m,h;
                           default: s)
  -j,--num-threads NUMBER  Number of threads to use for tests execution
                           (default: # of cores/capabilities)
  ...

Options for 'ListTestReporter':
  -l,--list-tests          Do not run the tests; just print their names

Options for 'tasty-quickcheck':
  --quickcheck-replay SEED Random seed to use for replaying a previous test run
                           (use same --quickcheck-max-size)
  ...
```
(As a side note, it'd be cool to have `PATTERN`s just be plain positional arguments)

[...]

Issue #341, opened 2022-07-11
Allow disabling / configuring test timing info


brandonchinn178:
Currently, if a test takes longer than a hardcoded limit, it's printed out.

https://github.com/UnkindPartition/tasty/blob/9b8ce7eb12046ce925b40ff2feb8b3ffde771c8d/core/Test/Tasty/Ingredients/ConsoleReporter.hs#L145-L146

It would be nice to be able to configure this limit. It would also be nice to never print the timing info (equivalent to limit = Infinity) or always print the timing info (equivalent to limit = -Infinity), but just configuring the limit would be sufficient for me.

Context: I'm writing integration tests that check the output of tasty, and currently, I have to strip out the timing information. https://github.com/brandonchinn178/tasty-autocollect/pull/7

Bodigrim:
Cf. https://github.com/UnkindPartition/tasty/pull/328

bfrk:
Yes, this is better than merely turning it on or off.

brandonchinn178:
No idea 🫤 Maybe something like `--min-duration-report`?

brandonchinn178:
`--min-duration-to-report` is better, just longer.

Sure, I might have some time to do this

PR #311, opened 2021-09-12
Tasty progress reporting

coot:
This PR is an updated and cleaned version of #245.

The upstream version introduced a bug which prevented quickcheck properties to
be shrinked.  The approach in this PR avoids that by using QuickCheck
facilities to print its own progress using tasty progress callback.

- Progress reporting ability.
- quickcheck: print progress
- Added  --no-progress option
- clean code

PR #343, opened 2022-07-19
Add exact dependency matching

martijnbastiaan:
Tasty 1.2 introduced a way for tests to specify dependencies. That is,
what tests should run before themselves. These dependencies are
specified using an AWK-like expression annotated to a `TestTree`. These
expressions can match against any test in the full `TestTree`. This
approach has a few rough edges:

  * Any pattern has to be tested against any test in the tree. If your
    test tree mostly consists of tests specifying dependencies for each
    other, this means calculating your test tree is of quadratic
    complexity.

  * It's easy to introduce dependency cycles, or other mistakes
    introducing needless sequentiality. The latter being especially
    insidious as it easily goes unnoticed.

This commit introduces the ability to specify dependencies by using a
combinator taking a list of `TestTree`s. This way of specifying dependencies
removes the quadratic complexity in favor of a linear one, while
eliminating the ability to accidentally introduce cycles or unintended
sequentiality.

-------------------

A bit of history for this PR as it has undergone several rewrites thanks
to a few great reviews I've gotten:

 * The PR originated from friction experienced over at `clash-lang/clash-compiler`
   where it was hard to define tests sequentially in a programmatic manner and
   startup of the testsuite now takes between 30-60 seconds. See [this comment](https://github.com/UnkindPartition/tasty/pull/343#issuecomment-1197768617).

 * The initial version added an `AfterTree TestTree TestTree` constructor, and
   would build a unique name for each test in the tree. Using a `Trie` it would
   then be able to relatively cheaply find dependencies of test cases. As a
   bonus, it would solve 3 out of 4 issues [documented in the README](https://github.com/UnkindPartition/tasty/pull/343#issuecomment-1195907083)
   for dependencies defined using patterns too. However, the implementation
   complexity of this was deemed to high to be merged.

 * The current implementation does not try to fix any problems with the existing
   way of defining dependencies, but instead focuses on just doing that for the
   new way. The new way ~~being an extra field added to `TestGroup`, which indicates
   whether tests should run in parallel or sequentially. To implement this
   properly, two backwards incompatible changes had to be made as documented
   [over here](https://github.com/UnkindPartition/tasty/pull/343#issuecomment-1229428427).~~ marks test trees as sequential using `IsOption ExecutionMode`. In
   order to implement sequential execution `foldGroup` had its type signature
   changed from `foldGroup :: OptionSet -> TestName -> b -> b` to 
   `foldGroup :: OptionSet -> TestName -> [b] -> b`, implemented in
   https://github.com/UnkindPartition/tasty/pull/364.

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -8,6 +8,7 @@ _YYYY-MM-DD_
 
 * Dependency loop error now lists all test cases that formed a cycle
 * `foldGroup` now takes `[b]` instead of `b` as its last argument to allow for custom fold strategies. This is a backwards incompatible change, but you can get the old behavior by applying `mconcat`.
+* Dependencies can now be defined pattern-free with `sequentialTestGroup`. (#343)
 
 Version 1.4.3
 ---------------
--- a/core/Test/Tasty/Core.hs
+++ b/core/Test/Tasty/Core.hs
@@ -16,8 +16,10 @@ module Test.Tasty.Core
   , ResourceSpec(..)
   , ResourceError(..)
   , DependencyType(..)
+  , ExecutionMode(..)
   , TestTree(..)
   , testGroup
+  , sequentialTestGroup
   , after
   , after_
   , TreeFold(..)
@@ -29,15 +31,21 @@ module Test.Tasty.Core
 
 import Control.Exception
 import qualified Data.Map as Map
+import Data.Bifunctor (Bifunctor(second, bimap))
+import Data.List (mapAccumR)
+import Data.Monoid (Any (getAny, Any))
+import Data.Sequence ((|>))
 import qualified Data.Sequence as Seq
 import Data.Tagged
 import Data.Typeable
 import GHC.Generics
+import Options.Applicative (internal)
 import Test.Tasty.Options
 import Test.Tasty.Patterns
 import Test.Tasty.Patterns.Types
 import Test.Tasty.Providers.ConsoleFormat
 import Text.Printf
+import Text.Read (readMaybe)
 
 -- | If a test failed, 'FailureReason' describes why.
 --
@@ -243,7 +251,28 @@ data DependencyType
   | AllFinish
     -- ^ The current test tree will be executed after its dependencies finish,
     -- regardless of whether they succeed or not.
-  deriving (Eq, Show)
+  deriving
+    ( Eq
+    , Show
+    , Read -- ^ @since 1.5
+    )
+
+-- | Determines mode of execution of a 'TestGroup'
+data ExecutionMode
+  = Sequential DependencyType
+  -- ^ Execute tests one after another
+  | Parallel
+  -- ^ Execute tests in parallel
+  deriving (Show, Read)
+
+-- | Determines mode of execution of a 'TestGroup'. Note that this option is
+-- not exposed as a command line argument.
+instance IsOption ExecutionMode where
+  defaultValue = Parallel
+  parseValue = readMaybe
+  optionName = Tagged "execution-mode"
+  optionHelp = Tagged "Whether to execute tests sequentially or in parallel"
+  optionCLParser = mkOptionCLParser internal
 
 -- | The main data structure defining a test suite.
 --
@@ -281,11 +310,20 @@ data TestTree
     --
     -- @since 1.2
 
--- | Create a named group of test cases or other groups
+-- | Create a named group of test cases or other groups. Tests are executed in
+-- parallel. For sequential execution, see 'sequentialTestGroup'.
 --
 -- @since 0.1
 testGroup :: TestName -> [TestTree] -> TestTree
 testGroup = TestGroup
+
+-- | Create a named group of test cases or other groups. Tests are executed in
+-- order. For parallel execution, see 'testGroup'.
+sequentialTestGroup :: TestName -> DependencyType -> [TestTree] -> TestTree
+sequentialTestGroup nm depType = setSequential . TestGroup nm . map setParallel
+ where
+  setParallel = PlusTestOptions (setOption Parallel)
+  setSequential = PlusTestOptions (setOption (Sequential depType))
 
 -- | Like 'after', but accepts the pattern as a syntax tree instead
 -- of a string. Useful for generating a test tree programmatically.
@@ -398,6 +436,18 @@ trivialFold = TreeFold
   , foldAfter = \_ _ _ b -> b
   }
 
+
+-- | Indicates whether a test matched in an evaluated subtree. If no filter was
+-- used, tests always match.
+type TestMatched = Any
+
+-- | Used to force tests to be included, even if they would be filtered out by
+-- a user's filter. This is used to force dependencies of a test to run. For
+-- example, if test @A@ depends on test @B@ and test @A@ is selected to run, test
+-- @B@ will be forced to match. Note that this only applies to dependencies
+-- specified using 'sequentialTestGroup'.
+type ForceTestMatch = Any
+
 -- | Fold a test tree into a single value.
 --
 -- The fold result type should be a monoid. This is used to fold multiple
@@ -439,7 +489,7 @@ foldTestTree0
      -- ^ the tree to fold
   -> b
 foldTestTree0 empty (TreeFold fTest fGroup fResource fAfter) opts0 tree0 =
-  go (filterByPattern (evaluateOptions opts0 tree0))
+  go (filterByPattern (annotatePath (evaluateOptions opts0 tree0)))
   where
     go :: AnnTestTree OptionSet -> b
     go = \case
@@ -478,23 +528,62 @@ evaluateOptions opts = \case
   After deptype dep tree ->
     AnnAfter opts deptype dep $ evaluateOptions opts tree
 
--- | Filter test tree by pattern, replacing leafs with 'AnnEmptyTestTree'.
-filterByPattern :: AnnTestTree OptionSet -> AnnTestTree OptionSet
-filterByPattern = go mempty
+-- | Annotate 'AnnTestTree' with paths.
+annotatePath :: AnnTestTree OptionSet -> AnnTestTree (OptionSet, Path)
+annotatePath = go mempty
   where
-    go :: Seq.Seq TestName -> AnnTestTree OptionSet -> AnnTestTree OptionSet
+    go :: Seq.Seq TestName -> AnnTestTree OptionSet -> AnnTestTree (OptionSet, Path)
     go path = \case
       AnnEmptyTestTree -> AnnEmptyTestTree
-      t@(AnnSingleTest opts name _)
-        | testPatternMatches (lookupOption opts) (path Seq.|> name)
-          -> t
-        | otherwise -> AnnEmptyTestTree
+      AnnSingleTest opts name tree -> 
+        AnnSingleTest (opts, path |> name) name tree
       AnnTestGroup opts name trees ->
-        AnnTestGroup opts name $ map (go (path Seq.|> name)) trees
+        let newPath = path |> name in
+        AnnTestGroup (opts, newPath) name (map (go newPath) trees)
       AnnWithResource opts res0 tree ->
-        AnnWithResource opts res0 $ \res -> go path (tree res)
+        AnnWithResource (opts, path) res0 $ \res -> go path (tree res)
       AnnAfter opts deptype dep tree ->
-        AnnAfter opts deptype dep (go path tree)
+        AnnAfter (opts, path) deptype dep (go path tree)
+
+-- | Filter test tree by pattern, replacing leafs with 'AnnEmptyTestTree'.
+filterByPattern :: AnnTestTree (OptionSet, Path) -> AnnTestTree OptionSet
+filterByPattern = snd . go (Any False)
+  where
+    go 
+      :: ForceTestMatch
+      -> AnnTestTree (OptionSet, Path)
+      -> (TestMatched, AnnTestTree OptionSet)
+    go forceMatch = \case
+      AnnEmptyTestTree ->
+        (Any False, AnnEmptyTestTree)
+
+      AnnSingleTest (opts, path) name tree
+        | getAny forceMatch || testPatternMatches (lookupOption opts) path
+        -> (Any True, AnnSingleTest opts name tree)
+        | otherwise 
+        -> (Any False, AnnEmptyTestTree)
+
+      AnnTestGroup (opts, _) name trees ->
+        case lookupOption opts of
+          Parallel ->
+            bimap
+              mconcat
+              (AnnTestGroup opts name)
+              (unzip (map (go forceMatch) trees))
+          Sequential _ ->
+            second
+              (AnnTestGroup opts name)
+              (mapAccumR go forceMatch trees)
+
+      AnnWithResource (opts, _) res0 tree ->
+        ( fst (go forceMatch (tree (throwIO NotRunningTests)))
+        , AnnWithResource opts res0 $ \res -> snd (go forceMatch (tree res))
+        )
+
+      AnnAfter (opts, _) deptype dep tree ->
+        second
+          (AnnAfter opts deptype dep)
+          (go forceMatch tree)
 
 -- | Get the list of options that are relevant for a given test tree
 treeOptions :: TestTree -> [OptionDescription]
--- a/core/Test/Tasty/Run.hs
+++ b/core/Test/Tasty/Run.hs
@@ -1,7 +1,7 @@
 -- | Running tests
 {-# LANGUAGE ScopedTypeVariables, ExistentialQuantification, RankNTypes,
              FlexibleContexts, CPP, DeriveDataTypeable, LambdaCase,
-             RecordWildCards #-}
+             RecordWildCards, NamedFieldPuns #-}
 module Test.Tasty.Run
   ( Status(..)
   , StatusMap
@@ -16,7 +16,7 @@ import Data.Int (Int64)
 import Data.Maybe
 import Data.List (intercalate)
 import Data.Graph (SCC(..), stronglyConnComp)
-import Data.Sequence (Seq, (|>), (<|))
+import Data.Sequence (Seq, (|>), (<|), (><))
 import Data.Typeable
 import Control.Monad (forever, guard, join, liftM)
 import Control.Monad.IO.Class (liftIO)
@@ -31,6 +31,10 @@ import Control.Arrow
 import Data.Monoid (First(..))
 import GHC.Conc (labelThread)
 import Prelude  -- Silence AMP and FTP import warnings
+
+#if MIN_VERSION_base(4,18,0)
+import Data.Traversable (mapAccumM)
+#endif
 
 #ifdef MIN_VERSION_unbounded_delays
 import Control.Concurrent.Timeout (timeout)
@@ -222,11 +226,8 @@ executeTest action statusVar timeoutOpt inits fins = mask $ \restore -> do
     -- See also https://github.com/UnkindPartition/tasty/issues/33
     yieldProgress _ = return ()
 
--- | Dependencies of a test
-type Dep = (DependencyType, Expr)
-
 -- | Traversal type used in 'createTestActions'
-type Tr = ReaderT (Path, [Dep]) IO (TestActionTree UnresolvedAction)
+type Tr = ReaderT (Path, Seq Dependency) IO (TestActionTree UnresolvedAction)
 
 -- | Exceptions related to dependencies between tests.
 --
@@ -252,13 +253,51 @@ instance Show DependencyException where
 
 instance Exception DependencyException
 
+-- | Specifies how to calculate a dependency
+data DependencySpec
+  = ExactDep (Seq TestName) (TVar Status)
+  -- ^ Dependency specified by 'TestGroup'. Note that the first field is only
+  -- there for dependency cycle detection - which can be introduced by using
+  -- 'PatternDep'.
+  | PatternDep Expr
+  -- ^ All tests matching this 'Expr' should be considered dependencies
+  deriving (Eq)
+
+instance Show DependencySpec where
+  show (PatternDep dep) = "PatternDep (" ++ show dep ++ ")"
+  show (ExactDep testName _) = "ExactDep (" ++ show testName ++ ") (<TVar>)"
+
+-- | Dependency of a test. Either it points to an exact path it depends on, or
+-- contains a pattern that should be tested against all tests in a 'TestTree'.
+data Dependency = Dependency DependencyType DependencySpec
+  deriving (Eq, Show)
+
+-- | Is given 'Dependency' a dependency that was introduced with 'After'?
+isPatternDependency :: Dependency -> Bool
+isPatternDependency (Dependency _ (PatternDep {})) = True
+isPatternDependency _ = False
+
+#if !MIN_VERSION_base(4,18,0)
+-- The mapAccumM function behaves like a combination of mapM and mapAccumL that
+-- traverses the structure while evaluating the actions and passing an accumulating
+-- parameter from left to right. It returns a final value of this accumulator
+-- together with the new structure. The accummulator is often used for caching the
+-- intermediate results of a computation.
+mapAccumM :: Monad m => (acc -> x -> m (acc, y)) -> acc -> [x] -> m (acc, [y])
+mapAccumM _ acc [] = return (acc, [])
+mapAccumM f acc (x:xs) = do
+  (acc', y) <- f acc x
+  (acc'', ys) <- mapAccumM f acc' xs
+  return (acc'', y:ys)
+#endif
+
 -- | An action with meta information
 data TestAction act = TestAction
   { testAction :: act
     -- ^ Some action, typically 'UnresolvedAction', 'ResolvedAction', or 'Action'.
   , testPath :: Path
     -- ^ Path pointing to this action (a series of group names + a test name)
-  , testDeps :: [Dep]
+  , testDeps :: Seq Dependency
     -- ^ Dependencies introduced by AWK-like patterns
   , testStatus :: TVar Status
     -- ^ Status var that can be used to monitor test progress
@@ -328,7 +367,7 @@ createTestActions opts0 tree = do
   -- to run them, and meta information needed to watch test progress and calculate
   -- dependencies in 'resolveDeps'.
   unresolvedTestTree :: TestActionTree UnresolvedAction <-
-    flip runReaderT (mempty :: (Path, [Dep])) $
+    flip runReaderT (mempty :: (Path, Seq Dependency)) $
       foldTestTree0 (pure (tGroup [])) (TreeFold { .. }) opts0 tree
 
   let
@@ -364,10 +403,16 @@ createTestActions opts0 tree = do
       pure $ TResource ini fin testTree
 
     foldAfter :: OptionSet -> DependencyType -> Expr -> Tr -> Tr
-    foldAfter _opts depType pat = local (second ((depType, pat):))
+    foldAfter _opts depType pat = local (second (Dependency depType (PatternDep pat) <|))
 
     foldGroup :: OptionSet -> TestName -> [Tr] -> Tr
-    foldGroup _opts name trees = tGroup <$> local (first (|> name)) (sequence trees)
+    foldGroup opts name trees =
+      fmap tGroup $ local (first (|> name)) $
+        case lookupOption opts of
+          Parallel ->
+            sequence trees
+          Sequential depType ->
+            snd <$> mapAccumM (goSeqGroup depType) mempty trees
 
     -- * Utility functions
     collectTests :: TestActionTree act -> [TestAction act]
@@ -382,25 +427,35 @@ createTestActions opts0 tree = do
       TGroup _ trees    -> mconcat (map collectFinalizers trees)
       TAction _         -> mempty
 
+    goSeqGroup 
+      :: DependencyType
+      -> Seq Dependency
+      -> Tr
+      -> ReaderT (Path, Seq Dependency) IO (Seq Dependency, TestActionTree UnresolvedAction)
+    goSeqGroup depType prevDeps treeM = do
+      tree0 <- local (second (prevDeps ><)) treeM
+
+      let
+        toDep TestAction {..} = Dependency depType (ExactDep testPath testStatus)
+        deps0 = Seq.fromList (toDep <$> collectTests tree0)
+
+        -- If this test tree is empty (either due to it being actually empty, or due
+        -- to all tests being filtered) we need to propagate the previous dependencies.
+        deps1 = if Seq.null deps0 then prevDeps else deps0
+
+      pure (deps1, tree0)
+
 -- | Take care of the dependencies.
 --
 -- Return 'Left' if there is a dependency cycle, containing the detected cycles.
 resolveDeps
   :: [TestAction ResolvedAction]
   -> Either [[Path]] [TestAction Action]
-resolveDeps tests = checkCycles $ do
+resolveDeps tests = maybeCheckCycles $ do
   TestAction { testAction=run_test, .. } <- tests
 
   let
-    -- Note: Duplicate dependencies may arise if the same test name matches
-    -- multiple patterns. It's not clear that removing them is worth the
-    -- trouble; might consider this in the future.
-    deps' :: [(DependencyType, TVar Status, Path)]
-    deps' = do
-      (deptype, depexpr) <- testDeps
-      TestAction { testStatus = testStatus1, testPath = testPath1 } <- tests
-      guard $ exprMatches depexpr testPath1
-      return (deptype, testStatus1, testPath1)
+    deps' = concatMap findDeps testDeps
 
     getStatus :: STM ActionStatus
     getStatus = foldr
@@ -429,7 +484,30 @@ resolveDeps tests = checkCycles $ do
           }
       }
   return (TestAction { testAction = action, .. }, (testPath, dep_paths))
+ where
+  -- Skip cycle checking if no patterns are used: sequential test groups can't
+  -- introduce cycles on their own.
+  maybeCheckCycles
+    | any (any isPatternDependency . testDeps) tests = checkCycles
+    | otherwise = Right . map fst
 
+  findDeps :: Dependency -> [(DependencyType, TVar Status, Seq TestName)]
+  findDeps (Dependency depType depSpec) =
+    case depSpec of
+      ExactDep testPath statusVar ->
+        -- A dependency defined using 'TestGroup' has already been pinpointed
+        -- to its 'statusVar' in 'createTestActions'.
+        [(depType, statusVar, testPath)]
+      PatternDep expr -> do
+        -- A dependency defined using patterns needs to scan the whole test
+        -- tree for matching tests.
+        TestAction{testPath, testStatus} <- tests
+        guard $ exprMatches expr testPath
+        [(depType, testStatus, testPath)]
+
+-- | Check a graph, given as an adjacency list, for cycles. Return 'Left' if the
+-- graph contained cycles, or return all nodes in the graph as a 'Right' if it
+-- didn't.
 checkCycles :: Ord b => [(a, (b, [b]))] -> Either [[b]] [a]
 checkCycles tests = do
   let
--- a/README.md
+++ b/README.md
@@ -691,13 +691,16 @@ Tasty executes tests in parallel to make them finish faster.
 If this parallelism is not desirable, you can declare *dependencies* between
 tests, so that one test will not start until certain other tests finish.
 
-Dependencies are declared using the `after` combinator:
+Dependencies are declared using the `after` or `sequentialTestGroup` combinator:
 
 * `after AllFinish "pattern" my_tests` will execute the test tree `my_tests` only after all
     tests that match the pattern finish.
 * `after AllSucceed "pattern" my_tests` will execute the test tree `my_tests` only after all
     tests that match the pattern finish **and** only if they all succeed. If at
     least one dependency fails, then `my_tests` will be skipped.
+* `sequentialTestGroup groupName dependencyType [tree1, tree2, ..]` will execute all tests
+   in `tree1` first, after which it will execute all tests in `tree2`, and so forth. Like
+   `after`, `dependencyType` can either be set to `AllFinish` or `AllSucceed`.
 
 The relevant types are:
 
@@ -707,6 +710,12 @@ after
   -> String         -- ^ the pattern
   -> TestTree       -- ^ the subtree that depends on other tests
   -> TestTree       -- ^ the subtree annotated with dependency information
+
+sequentialTestGroup
+  :: TestName       -- ^ name of the group
+  -> DependencyType -- ^ whether to run the tests even if some of the dependencies fail
+  -> [TestTree]     -- ^ trees to execute sequentially
+  -> TestTree
 
 data DependencyType = AllSucceed | AllFinish
 ```
@@ -744,7 +753,7 @@ tests. The resource may or may not be managed by `withResource`.)
      ]
    ```
 
-Here are some caveats to keep in mind regarding dependencies in Tasty:
+Here are some caveats to keep in mind when using patterns to specify dependencies in Tasty:
 
 1. If Test B depends on Test A, remember that either of them may be filtered out
    using the `--pattern` option. Collecting the dependency info happens *after*
@@ -770,6 +779,8 @@ Here are some caveats to keep in mind regarding dependencies in Tasty:
    dependencies is sufficiently different from the natural order of tests in the
    test tree, searching for the next test to execute may also have an
    overhead quadratic in the number of tests.
+
+Use `sequentialTestGroup` to mitigate these problems.
 
 
 ## FAQ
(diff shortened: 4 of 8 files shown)


martijnbastiaan:
@Bodigrim @VictorCMiraldo I'm eager to get this PR merged, my aim is to fix all of the gotchas listed in Tasty's README regarding dependencies - these affect the main project I'm working on:

> 1. If Test B depends on Test A, remember that either of them may be filtered out
>    using the `--pattern` option. Collecting the dependency info happens *after*
>    filtering. Therefore, if Test A is filtered out, Test B will run
>    unconditionally, and if Test B is filtered out, it simply won't run.
> 1. Tasty does not currently check whether the pattern in a dependency matches
>    anything at all, so make sure your patterns are correct and do not contain
>    typos. Fortunately, misspecified dependencies usually lead to test failures
>    and so can be detected that way.
> 1. Dependencies shouldn't form a cycle, otherwise Tasty with fail with the
>    message "Test dependencies have cycles." A common cause of this is a test
>    matching its own dependency pattern.
> 1. Using dependencies may introduce quadratic complexity. Specifically,
>    resolving dependencies is *O(number_of_tests × number_of_dependencies)*,
>    since each pattern has to be matched against each test name. As a guideline,
>    if you have up to 1000 tests, the overhead will be negligible, but if you
>    have thousands of tests or more, then you probably shouldn't have more than a
>    few dependencies.
> 
>    Additionally, it is recommended that the dependencies follow the
>    natural order of tests, i.e. that the later tests in the test tree depend on
>    the earlier ones and not vice versa. If the execution order mandated by the
>    dependencies is sufficiently different from the natural order of tests in the
>    test tree, searching for the next test to execute may also have an
>    overhead quadratic in the number of tests.

This PR together with [a followup PR](https://github.com/martijnbastiaan/tasty/pull/3) should achieve that. I know it's quite a bit to review, so I've tried to add enough tests to hopefully convince you that it's at least somewhat sound. If you'd like a video call to talk this through, I'm open for that too! If there's anything else I can do, ask away.

Bodigrim:
@martijnbastiaan sorry, I don't have bandwidth for this right now. Could you possibly summarise all breaking changes here (both API and behaviour)?

CC @VictorCMiraldo @adamgundry @andreasabel as other maintainers.

martijnbastiaan:
No worries @Bodigrim, thanks for the reply.

The API changes are as follows:

 * Added exports: `afterTree`, `sequentialTestGroup`.
 * Added `TreeFold.foldAfterTree`. This allows folding `AfterTree` (a constructor not part of the public API).
 * `TreeFold.foldSingle` takes an extra argument `ExactPath`. I would actually like every function in `TreeFold` to take an `ExactPath` (its position in the `TestTree`), but I decided against is to minimize API changes. Given its usefulness I'd like to ask the maintainers here to make the same consideration - and let me know if they'd opt for adding the `ExactPath` everywhere.

There are no behavioral changes to existing programs.

VictorCMiraldo:
Thanks for the PR @martijnbastiaan! That is a large one! :) 
Unfortunately, I don't have the bandwidth nor the familiarity with the mechanisms surrounding `tasty`'s test dependencies. Hence, I'm not sure I can give a good review on this anytime soon.

martijnbastiaan:
Understood! 

Hmm. So to give some background: I wrote this patch to resolve some problems we face with our 4K large testsuite over at [clash-lang/clash-compiler](https://github.com/clash-lang/clash-compiler/), a project me and my colleagues at QBayLogic maintain. 

What I could do is ask a colleague for a review on this patch (and the follow-up one). Would that be enough -together with the listed API changes I posted earlier- to accept this patch? I'm of course willing to maintain the feature too, and deal with the fallout if any.

Bodigrim:
> TreeFold.foldSingle takes an extra argument ExactPath. 

Can this breaking change be avoided by introducing a new fold and expressing existing foldSingle via it?

martijnbastiaan:
Yes, we could add `foldSingleExact` that will (by default) simply call `foldSingle`.

core/Test/Tasty/Core.hs
```diff
@@ -303,16 +308,41 @@ after deptype s =
...
+  go (t:ts) = Just (foldl' (afterTree depType) t ts)
+
 -- | An algebra for folding a `TestTree`.
 --
 -- Instead of constructing fresh records, build upon `trivialFold`
 -- instead. This way your code won't break when new nodes/fields are
 -- indroduced.
 data TreeFold b = TreeFold
-  { foldSingle :: forall t . IsTest t => OptionSet -> TestName -> t -> b
+  { foldSingle :: forall t . IsTest t => OptionSet -> ExactPath -> TestName -> t -> b
```
Bodigrim, line 341: Ah, I see. The breaking change is in a record field. This is nasty, unlikely one can hide it beneath `PatternSynonyms`. 

Changing `TreeFold` is likely to break all third-party `tasty` ingredients.

Bodigrim:
@martijnbastiaan could you please share what kind of dependency graph is typical for Clash test suite? Is it chains of depending tests?

martijnbastiaan:
Sure. To give a bit of context: Clash is a Haskell to silicon compiler - meaning Haskell can be used to build chips that could be synthesized by say a TSMC. Silicon is typically written in a Hardware Description Language (HDL) which is what Clash targets as a compiler output. There are currently three major HDLs: Verilog, SystemVerilog, and VHDL. Each of these languages have various simulators that can be used to predict chip behavior before actually burning it into silicon. Clash itself can also be simulate (simply by leveraging GHC!).

So, one test looks something like:

```mermaid
  graph TD;
      FooTest-->VHDL;
      FooTest-->Verilog;
      FooTest-->SystemVerilog;
      VHDL-->Clash_VHDL;
      VHDL-->Clash_Sim_VHDL;
      Verilog-->Clash_Verilog;
      Verilog-->Clash_Sim_Verilog;
      SystemVerilog-->Clash_SV;
      SystemVerilog-->Clash_Sim_SV;
      Clash_SV-->ModelSim;
      Clash_SV-->Verilator;
      Clash_VHDL-->GHDL;
      Clash_VHDL-->Vivado_VDHL;
      Clash_Verilog-->Vivado_Verilog;
      Clash_Verilog-->IcarusVerilog;
```

I'm glossing over some things, but each new (regression) test looks a bit like this. This makes the number of dependencies grow linearly with the number of tests, hence us running into this caveat mentioned in the README:

> Using dependencies may introduce quadratic complexity. Specifically, resolving dependencies is O(number_of_tests × number_of_dependencies), since each pattern has to be matched against each test name. As a guideline, if you have up to 1000 tests, the overhead will be negligible, but if you have thousands of tests or more, then you probably shouldn't have more than a few dependencies.

With regards to:

> Changing TreeFold is likely to break all third-party tasty ingredients.

I'm willing to help out patching them, of course.

core/Test/Tasty/Core.hs
martijnbastiaan: It would. I think we can make this backwards compatible by defining `foldSingleExact` that takes the original `foldSingle` as an argument.

But... I don't actually think it's worth it. Patching any existing code would be a matter of ignoring one (more) argument / fairly mechanical, which seems like a fair trade-off IMO.

Bodigrim:
I agree that the infrastructure for dependencies between tests is wanting, but I feel that your approach covers only a fraction of cases, but required changes are quite profound. My use case is dependencies between disjoint, potentially remote trees, see https://hackage.haskell.org/package/tasty-bench-0.3.2/candidate/docs/Test-Tasty-Bench.html#g:4

If the shape of your dependency graphs is relatively stable, might your use case be better served with a custom `IsTest` provider?

martijnbastiaan:
The only real limitation is that you can't make a DAG, you can only make a tree. I've checked out the example of [bcompare](https://hackage.haskell.org/package/tasty-bench-0.3.2/candidate/docs/Test-Tasty-Bench.html#v:bcompare) and the examples it links (`chimera`, `fast-digits`, `unicode-data`). Correct me if I'm wrong, but it seems like none of those examples need the `TestTree` to be a DAG. I think all of these examples just want to run a single test as a baseline (left part of `afterTree`) and compare it against a bunch of other tests (right part of `afterTree`). I'll respond to the other points too, but I'd like to understand the need for a DAG better before I do.

Bodigrim:
Correct, none of my examples needs a DAG. However, the shape of `TestTree` is not necessarily in correspondence with a dependency tree.

martijnbastiaan:
Sure, that's true. I'll be honest, it's kinda hard to respond to this without sounding very negative/argumentative. It seems to me that you're paying a pretty high price for the ability to do so. The examples need to construct patterns such as:

```haskell
"$NF == \"" ++ show n ++ "\" && $(NF-1) == \"Chimera\""
```

..which would break if there's ever another test group called `Chimera` _anywhere_ in the tree, quite seriously contradicting the composability of `TestTree`s. It has all the downsides mentioned in the README. Last but not least, for a casual reader it's pretty hard to even read the code - given that you need to do the same process Tasty does when finding a dependency: scan the whole tree.

So all in all, it seems somewhat "obvious" to me that you'd want something like `afterTree`. And as far as the complexity goes.. well that's kinda imposed to us by the existence of `after` itself. If we _only_ had `afterTree`, there would have been no need for `Trie` or `DependencyTree` (from the follow-up PR). All would naturally follow from the trees construction and traversal - the code would be even simple than what's now in `main`.

So, in a way I feel that this remark:

> but I feel that your approach covers only a fraction of cases

turns everything on its head. It seems to me that `after` should try and justify its costs. Is it really worth having this pattern matching when it messes with composability, computational complexity, correct-by-constructionness, and code complexity?

I guess what I'm asking is: is there any chance of this getting merged? If not, do you see an alternative way of tackling the issues in the README, or will Tasty be forever stuck with them?

Bodigrim:
@martijnbastiaan to be very clear, I agree that the dependency design is inconvenient and I also agree that if we were doing it from the scratch and with our current experience, then `AfterTree` would be a saner option. My problem is that we already have a mechanism (constructor `After`) for this pretty niche feature, and adding another one (constructor `AfterTree`) for even more niche use cases does not strike me as the most desirable outcome. I'm happy to be overruled by other maintainers, my cautiousness is because I'm a caretaker only. 

So, answering your question directly, yes, there is a chance, but we must consider the design and long-term consequences. This can take more time than one might like, sorry for that. I do appreciate your work, your efforts and this particular contribution.

What is the prior art for dependencies between tests in other frameworks?

----

Could you define something like 

```haskell
data ClashTest = ClashTest 
  { ghdlTest :: Assertion
  , vivadoVdhlTest :: Assertion
  , vivadoVerilogTest :: Maybe Assertion
  , icarusVerilogTest :: [Assertion]
  , ...
  }
```

and provide `instance IsTest ClashTest` to manage dependency chain? More generally, you can define `newtype AssertionChain = AssertionChain [Assertion]` with `instance IsTest AssertionChain` executing tests in order.

martijnbastiaan:
Understood. If necessary, I'd be willing to maintain it. My ideal would be to deprecate `After` in the very long term, but I understand that takes a lot of time, effort, and commitment - hence not always worth it.

I'll have a look at your proposal and dig into the prior art, hopefully this weekend.

martijnbastiaan:
> What is the prior art for dependencies between tests in other frameworks?

I've looked into this:

 * [sydtest](https://hackage.haskell.org/package/sydtest): exposes sequential and parallel declaring that all tests in a test tree need to be executed sequentially or in parallel. (Where parallel is the default.) It does not seem to offer a way to chain two tests together.
 * [hspec](https://hackage.haskell.org/package/hspec): exposes parallel declaring all tests in a test tree to be executed in parallel. It does not seem to offer a way to modify this locally - i.e. further down in the test case.

To me it looks like sydtest is very close to doing TheRightThing(tm).

> Could you define something like

This can work, _but_ it has quite a few drawbacks by virtue of the tests being opaque to Tasty:

 * It's not rendering them in the progress report, also making it harder to read test failures
 * It's hard(er) to run specific simulators individually / replay tests
 * It defeats parallelism between different simulators

So all in all this would be a pretty big step back IMO.

------------------------------------

I could do two things to make this series of patches more acceptable: 

  * I could reverse the order of patches, making the first patch (fixing test filtering, [a PR](https://github.com/martijnbastiaan/tasty/pull/3) I intended to publish after merging this one) a pure bugfix. Having that infrastructure would make adding `AfterTree` almost trivial.
  * I could make the patches fully backwards compatible by introducing a bunch of "exact" functions, like:

```haskell
-- | An algebra for folding a `TestTree`.
--
-- Instead of constructing fresh records, build upon `trivialFold`
-- instead. This way your code won't break when new nodes/fields are
-- indroduced.
data TreeFold b = TreeFold
  { foldSingle :: forall t . IsTest t => OptionSet -> TestName -> t -> b
  , foldGroup :: OptionSet -> TestName -> b -> b
  , foldResource :: forall a . OptionSet -> ResourceSpec a -> (IO a -> b) -> b
  , foldAfter :: OptionSet -> DependencyType -> Expr -> b -> b

    -- "Exact" versions of functions defined above. In 'trivialFold', these simply
    -- call the non-exact versions. These functions are here for backwards compatiblity
    -- reasons.
    --
    -- Note: instead of taking a `TreeFold`, these functions could take their non-exact
    -- function directly, but this makes the type signatures quite long.
  , exactFoldSingle :: forall t . IsTest t => TreeFold b -> ExactPath -> OptionSet -> TestName -> t -> b
  , exactFoldGroup :: TreeFold b -> ExactPath -> OptionSet -> TestName -> b -> b
  , exactFoldResource :: forall a . TreeFold b -> ExactPath -> OptionSet -> ResourceSpec a -> (IO a -> b) -> b
  , exactFoldAfter :: TreeFold b -> ExactPath -> OptionSet -> DependencyType -> Expr -> b -> b
  }
```

I personally don't think it's worth it, but I'm happy to do it if that makes things more acceptable/reviewable.

----------

> I do appreciate your work, your efforts and this particular contribution.

Thanks btw, happy to work with you guys.

Bodigrim:
> To me it looks like sydtest is very close to doing TheRightThing(tm).

Would `sequential` combinator on its own be sufficient to match your requirements?

martijnbastiaan:
For my specific use case it would be both a step forward and backward at the same time.

The positives:

 * We could remove the code dealing with patterns, which is a fairly ugly/confusing part of the test code now.
 * Starting the testsuite would (if properly implemented) again be a matter of milliseconds, instead of half a minute.

The neutrals:

 * Tasty would continue to show tests separately.

The negatives:

 * We'd lose the ability to run simulators in parallel.

------------

I'm somewhat biased as I already put in the work of course, but it strikes me as a band-aid instead of a proper solution. Do you think it would be easier to implement?

Bodigrim:
Well, adding `TestTreeSequential` constructor or maybe extending existing `TextTree` with a switch `Parallel | Sequential` seems to me a better option. You do not lose ability to run certain parts in parallel, because `TestTreeSequential` can affect only one level of tree.

martijnbastiaan:
Sure, I can amend the patch to do this.

Bodigrim:
@martijnbastiaan before you do more work, I'd like to hear from other maintainers.

@VictorCMiraldo @andreasabel Existing `TestTree` constructor groups subtrees for (potentially) parallel execution. The proposal is either to add `TestTreeSequential` constructor, or to extend `TestTree` with a field `data Parallelism = Parallel | Sequential`, which mandates parallel execution of subtrees. How do you feel about this? Currently a similar (but not quite the same) effect can be achieved with a sequence of `After` constructors, which is very tedious to write and unbearably slow to execute.

core/Test/Tasty/Core.hs
```diff
@@ -303,16 +308,41 @@ after deptype s =
...
   , foldGroup :: OptionSet -> TestName -> b -> b
   , foldResource :: forall a . OptionSet -> ResourceSpec a -> (IO a -> b) -> b
   , foldAfter :: OptionSet -> DependencyType -> Expr -> b -> b
+  , foldAfterTree :: OptionSet -> DependencyType -> ExactPath -> b -> b
```
VictorCMiraldo, line 345: From the name and the type of `AfterTree`, I would have expected the type of `foldAfterTree` to be `OptionSet -> DependencyType -> ExactPath -> b -> b -> b`. Does this change mean that the definition of `foldTestTree` is not really a catamorphism anymore? (tbh, I'm not aware if it even ever was one!)

VictorCMiraldo:
@Bodigrim and @martijnbastiaan, first of all, thanks for the great work both making the PR and the thoughtful comments! :)

> Currently a similar (but not quite the same) effect can be achieved with a sequence of `After` constructors, which is very tedious to write and unbearably slow to execute.

Yeah, that doesn't sound like a very good solution and I can see how running tests sequentially is an interesting feature. I'm not sure how I feel about relying on the test dependency mechanism to impose a sequential execution order.
I do tend to agree with @Bodigrim that adding a `AfterTree` constructor feels weird given that we already have an `After`. Considering the semantic interplay between them seems like a potential nightmare for me. Can we have weird deadlocks with things like (in pseudo-tasty) `t = AfterTree (After "/a/" $ singleTest "b") (singleTest "a")`? IIUC, plain `tasty` already has this type of problem with regular `TestTree`s.

It does feel more natural to to add a `TestTreeSequential` constructor instead of using the dependency mechanism to enforce sequential execution. Also, it doesn't introduce new quirks w.r.t. dependencies such as `t` above:
a `TestTree u` has a deadlock iff `TestTreeSequential u` has a deadlock. 

Would a `TestTreeSequential` need [this change](https://github.com/UnkindPartition/tasty/pull/343/files/2b5c589ff8b257c36af10d3d7617d7e7d8541c65#diff-6f4da76a7d8d63450fc36ab4b935eb551df8a93bc1b1d784414118ee80be62a6R341)? If not, then it would be fully backwards compatible, which is a huge plus IMO

Bodigrim:
Potentially we can leave `TreeFold` as is, with `foldGroup` processing both `TestGroup` and `TestGroupSequential`. Adding a new constructor to `TestTree` is breaking any way, but potentially with less impact on clients.

martijnbastiaan:
Fundamentally, `TestTreeSequential` will differ very little from `AfterTree`. Instead of having two `TestTree` fields, it will simply take a list of them. I do think it's the better option: I only ended up using `sequentialTestGroup` (a function this PR introduces) in practice. However, it will still need the changes to `TreeFold`, unless I find a better way to implement it.

The core of the problem is that this variable:

https://github.com/UnkindPartition/tasty/blob/df7ccab50d361c7d1a3c00960faac883d98c2530/core/Test/Tasty/Run.hs#L327-L332

needs to take `TestTreeSequential` (or `AfterTree`) into account. While patterns need to scan the whole tree there, `AfterTree`/`TestTreeSequential` can define much more efficient lookup strategies.

core/Test/Tasty/Core.hs
martijnbastiaan: I ended up only using one, but I think it could take both. As for the catamorphism; I've tried to read up on it but I can't make heads or tails of it.

martijnbastiaan:
Hmm.. An alternative implementation could implement a custom folder that propagates dependencies up while it is folding. I.e., it would not use `foldTestTree` to run the tests. This would avoid the complexities of introducing the `Trie` and would also make for a natural solution to the very first caveat mentioned in the README. I'm not sure how this would fit into the rest of the ecosystem though.

core/Test/Tasty/Core.hs
Bodigrim: > Does this change mean that the definition of foldTestTree is not really a catamorphism anymore? (tbh, I'm not aware if it even ever was one!)

`foldTestTree` is not a catamorphism even now: there is no direct correspondence between constructors of `TestTree` and fields of `TreeFold`.

Bodigrim:
I would do this (please find a better name for `TestGroup'`):
```haskell
data ExecutionOrder = Concurrent | Sequential

data TestTree
  = forall t . IsTest t => SingleTest TestName t
  | TestGroup' ExecutionOrder TestName [TestTree]
  | PlusTestOptions (OptionSet -> OptionSet) TestTree
  | forall a . WithResource (ResourceSpec a) (IO a -> TestTree)
  | AskOptions (OptionSet -> TestTree)
  | After DependencyType Expr TestTree

pattern TestGroup = TestGroup' Concurrent
```
This is a breaking change, but a very minor one. And you can leave `TreeFold` as is, ignoring `ExecutionOrder`, so there is no breakage here. 

Indeed, `createTestActions` can fold `TestTree` by its own means, not necessarily via `foldTestTree`.

VictorCMiraldo:
FWIW, I like to compromise suggested by @Bodigrim above. Seems like a good option to minimize breakage while enabling @martijnbastiaan to have the necessary feature that motivated the PR in the first place.

martijnbastiaan:
Great, thank you both. I've got a patch now that implements the suggestions (adding TestGroup field + custom traversal). While it doesn't try to fix any of the issues raised in the README for pattern dependencies, it does solve them for the new way of specifying deps in a fairly straightforward way.

I'm pretty happy with the patch, but I'm still working on writing a proper testsuite. I'm spending only a few hours on it every weekend so it'll probably take a while..

martijnbastiaan:
I've just pushed an alternative implementation based on the comments above. This one doesn't try to fix any issues with dependencies specified using patterns, but does solve them all for dependencies specified using `sequentialTestGroup`. `TestGroup` is now the primitive setting parallelization/sequentiality. Two backwards incompatible changes have been applied:

----------------------------------------------

**1. `TreeFold` no longer has a field `foldResource`.** Its type signature had to change in order for the fold to account for dependencies while filtering. However, it doesn't seem likely downstream will be affected at all:

1. [Searching for `foldResource` on all of GitHub](https://github.com/search?l=Haskell&q=foldResource&type=Code) only yields `tasty` using it.
2. Searching through all libraries with `tasty` in their name on Hackage does not yield any relevant usages of `TreeFold` either.

If this change proves controversial, I suggest we move the second commit in this PR to a separate PR and discuss there.

----------------------------------------------

**2. `TestGroup` has an additional field `ExecutionMode`.** @Bodigrim, you suggested making a pattern synonym like this:

```haskell
pattern TestGroup = TestGroup' Concurrent
```

But, libraries using this pattern will silently fail:

```haskell
case tree of
  TestGroup _ -> ...
  _           -> ...
```

Libraries using the following pattern will fail with just a GHC warning:

```haskell
case tree of
  TestGroup _ -> ..
  [.. exhaustive list of constructors ..]
```

To get a feel for the breakage, I've cloned all the libraries published on Hackage related to `tasty` and searched for `TestGroup`. Of all these libraries, 7 use `TestGroup` directly: `tasty-bench`, `tasty-expected-failure`, `tasty-silver`, `tasty-inspection-testing`, `tasty-focus`, `tasty-focus`, and `tasty-bdd`. Of these, [6 would fail silently as their usage falls in one of the two patterns mentioned](https://gist.github.com/martijnbastiaan/0c348a46ddec1fd3f63134f7b17843cd). Only `tasty-inspection-testing` would continue to be correct.

I'm proposing to let static typing do its thing to prevent silent breakage.

core/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Unreleased
 _YYYY-MM-DD_
 
 - Dependency loop error now lists all test cases that formed a cycle
+- Dependencies can now be defined pattern-free with `sequentialTestGroup`
```
Bodigrim, line 10: Please describe in changelog, which entities were added and which has been changed.

core/Test/Tasty/Core.hs
```diff
@@ -330,10 +346,20 @@ trivialFold :: Monoid b => TreeFold b
 trivialFold = TreeFold
   { foldSingle = \_ _ _ -> mempty
   , foldGroup = \_ _ b -> b
-  , foldResource = \_ _ f -> f $ throwIO NotRunningTests
   , foldAfter = \_ _ _ b -> b
   }
 
+newtype Matched = Matched Bool
```
Bodigrim, line 352: What's wrong with `Any`?

core/Test/Tasty/Core.hs
```diff
@@ -311,7 +328,6 @@ after deptype s =
 data TreeFold b = TreeFold
   { foldSingle :: forall t . IsTest t => OptionSet -> TestName -> t -> b
   , foldGroup :: OptionSet -> TestName -> b -> b
```
Bodigrim, line 330: If we must break `TreeFold` anyway, it's probably better to change `foldGroup` to take `ExecutionMode` into account. Would it allow to express `createTestActions` via `foldTestTree`?

core/Test/Tasty/Run.hs
```diff
@@ -247,89 +234,158 @@ instance Show DependencyException where
...
+  , testAction  :: (Seq Initializer, Seq Finalizer) -> IO ()
+  , testStatus  :: TVar Status
+  , testDepends :: Seq dep
+  }
+
+-- Behaves like a combination of fmap and foldl; it applies a function to each
+-- element of a structure, passing an accumulating parameter from left to right,
+-- and returning a final value of this accumulator together with the new
+-- structure. Monadic version of 'mapAccumL'.
+mapAccumLM :: Monad m => (acc -> x -> m (acc, y)) -> acc -> [x] -> m (acc, [y])
```
Bodigrim, line 272: Please rename to `mapAccumM`, similar to `base-4.18` (https://gitlab.haskell.org/ghc/ghc/-/commit/4b98c5ce971b4ea6a2ca9e44f2d068088546751a).

core/Test/Tasty/Run.hs
```diff
@@ -247,89 +234,158 @@ instance Show DependencyException where
...
+    -> Seq Dependency
+    -> (Seq (TestAction Dependency), Seq Finalizer)
+    -> (Seq Dependency, (Seq (TestAction Dependency), Seq Finalizer))
+  goSeqGroup depType deps0 (actions0, finalizers) = do
+    (deps1, (actions1, finalizers))
+   where
+    actions1 = fmap updateDepends actions0
+
+    updateDepends a@TestAction{testDepends} =
+      a{testDepends=deps0 <> testDepends}
```
Bodigrim, line 365:
```suggestion
      a { testDepends = deps0 <> testDepends }
```

Bodigrim (approved):

core/Test/Tasty/Core.hs
martijnbastiaan: I don't think so. The current traversal in `Run.hs` not only propagates info up, it also propagates info down (`Seq Dependency`). I don't see an obvious way to achieve that with `foldTestTree` (unless it is radically changed).

core/Test/Tasty/Core.hs
martijnbastiaan: Nothing! I wasn't aware of its existence.

martijnbastiaan:
@Bodigrim I've applied your suggestions in separate commits. I'll squash them into the first two if you're happy with them.

core/Test/Tasty/Core.hs
martijnbastiaan: Hmm, perhaps it doesn't need to. I'll see if I can make a patch and document the consequences.

core/Test/Tasty/Core.hs
```diff
@@ -361,22 +377,31 @@ foldTestTree
...
      -- ^ the tree to fold
   -> b
-foldTestTree (TreeFold fTest fGroup fResource fAfter) opts0 tree0 =
-  go mempty opts0 tree0
+foldTestTree (TreeFold fTest fGroup fAfter) opts0 tree0 =
+  snd (go mempty opts0 mempty tree0)
   where
-    go :: (Seq.Seq TestName -> OptionSet -> TestTree -> b)
-    go path opts tree1 =
+    go :: Seq.Seq TestName -> OptionSet -> Matched -> TestTree -> (Matched, b)
```
Bodigrim, line 383: I'm struggling with this: who precisely has `Matched`? Probably a good place to attach a comment, explaining the algorightm.

martijnbastiaan: Will do, but perhaps this will clarify it in the mean time:

```haskell
type TestMatched = Any
type ForceMatch = Any

go :: Seq.Seq TestName -> OptionSet -> ForceFilter -> TestTree -> (TestMatched, b)
```

Basically, when test `A` depends on `B`, and `A` is matched with `-p A` - `B` should also run. The code therefore keeps track of any matched tests, and forces their dependencies to match too (even if the filter would otherwise throw them out).

core/Test/Tasty/Core.hs
```diff
@@ -361,22 +377,31 @@ foldTestTree
   -> TestTree
      -- ^ the tree to fold
   -> b
-foldTestTree (TreeFold fTest fGroup fResource fAfter) opts0 tree0 =
-  go mempty opts0 tree0
+foldTestTree (TreeFold fTest fGroup fAfter) opts0 tree0 =
+  snd (go mempty opts0 mempty tree0)
```
Bodigrim, line 381: It gets a bit convoluted. Can we decompose this function into two passes: one to apply pattern `TestPattern -> TestTree -> TestTree`, and another to actually fold?

core/Test/Tasty/Core.hs
```diff
@@ -361,22 +377,31 @@ foldTestTree
...
-        AskOptions f -> go path opts (f opts)
-        After deptype dep tree -> fAfter opts deptype dep $ go path opts tree
+        TestGroup Parallel name trees ->
+          second
+            (fGroup opts name)
+            (foldMap (go (path Seq.|> name) opts matched) trees)
+        TestGroup (Sequential _) name trees ->
+          second
+            (fGroup opts name . mconcat . reverse)
+            (mapAccumL (go (path Seq.|> name) opts) matched (reverse trees))
```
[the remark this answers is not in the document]
martijnbastiaan: Done. I've applied a similar change in the other traversal.

core/Test/Tasty/Core.hs
martijnbastiaan: Done

core/Test/Tasty/Core.hs
martijnbastiaan: I'm not sure what this would look like. Could you spell it out a bit more? If it helps, I've added inline documentation.

core/Test/Tasty/Core.hs
martijnbastiaan: I've played with this a bit more but I can't get it to work. Admittedly, this was a while ago so I don't exactly remember what didn't work out..

martijnbastiaan:
@Bodigrim Again thanks for all the comments, and sorry for the delay on this one. There are two unresolved comments now. I've been unable to come up with a strategy to join `foldTestTree` and the fold in `createTestActions` - I couldn't find an implementation/type for `foldResource` that worked out. As discussed before, I don't think anyone but Tasty itself uses this field, so I don't think this is a major issue.

martijnbastiaan:
@VictorCMiraldo @Bodigrim Anything I could do to get this PR going again? :)

Bodigrim:
@martijnbastiaan I'm sorry, I've lost all the context since August 2022 and I don't have time to restore it at the moment. I'll try to review in upcoming weeks.

martijnbastiaan:
Yeah no worries, my fault for being a slowpoke. I'll brush up the cover letter with an overview of the history of this PR today.

core/Test/Tasty/Core.hs
Bodigrim: See https://github.com/UnkindPartition/tasty/pull/363. I imagine you should be able to modify `filterByPattern` only.

Bodigrim:
@martijnbastiaan could you please rebase?

martijnbastiaan:
I managed to make it work without the custom traversal in `Run.sh` and with only one very minor API change. I need some more time to brush it up for review. I hope to publish that this weekend.

core/Test/Tasty/Core.hs
```diff
@@ -259,7 +273,7 @@ data DependencyType
 data TestTree
   = forall t . IsTest t => SingleTest TestName t
     -- ^ A single test of some particular type
-  | TestGroup TestName [TestTree]
+  | TestGroup ExecutionMode TestName [TestTree]
```
Bodigrim, line 276: Can we pass `ExecutionMode` as an `Option` rather than a dedicated field of `TestGroup`? This would avoid any breakage to `data TestTree` and is more `tasty`, because pretty much every other execution option is passed like this.

Bodigrim (commented):
Overall looks very nice. I'll take a more in-depth look to validate changes, in the meantime I have one architectural suggestion.

core/Test/Tasty/Core.hs
martijnbastiaan: Clever, I like it. Will do.

martijnbastiaan:
Thanks @Bodigrim. Regarding the commit:

> [Refactor 'createTestActions'](https://github.com/UnkindPartition/tasty/pull/343/commits/740eaa4eacc6717a7bb6069b573ea67dd6f9b076)

I found the stack of `Traversal`/`Reader`/`Writer`/`IO` + monoidal combination to be very cumbersome to work with. Given that it also had some complexity issue I decided to refactor it into the current structure (see the commit message for more information). Overall I'd say the current approach is an improvement, but if there's disagreement I'll try my best to make it work in the original structure.

If it helps; we could also split this PR up into multiple (cleanup + `foldGroup` API change + refactor `createTestActions` + this).

core/Test/Tasty/Core.hs
martijnbastiaan: Options propagate all the way down, but there's no reason execution mode information should. The only way around that seems that `testGroupSequential` would do something like this pseudocode:

```
TestGroup [t1, t2]

=>

setSequential (TestGroup "x" [setParallel t1, setParallel t2])
```

That would ensure the options of test group `x` would evaluate to "test sequentially" while evaluation the options for t1/t2 would yield "test in parallel".

This seems kinda weird/magical, do we want to pursue this?

Bodigrim: I think that's fine, yes.

martijnbastiaan: Ok, done! I'll squash the new commit into its parent if we intend to merge.

Bodigrim:
Let's pull out refactoring to its own PR please.

martijnbastiaan:
Okay, done: https://github.com/UnkindPartition/tasty/pull/364.

core/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ _YYYY-MM-DD_
 
 * Dependency loop error now lists all test cases that formed a cycle
 * `foldGroup` now takes `[b]` instead of `b` as its last argument to allow for custom fold strategies. This is a backwards incompatible change, but you can get the old behavior by applying `mconcat`.
+* Dependencies can now be defined pattern-free with `sequentialTestGroup`. (#343)
```
Bodigrim, line 11: What's the interaction of `--pattern` and dependencies? Is it unchanged? I was under impression that now if a test matches `--pattern`, its prerequisites are also executed, even if they don't match.

core/Test/Tasty/Core.hs
```diff
@@ -243,7 +250,21 @@ data DependencyType
   | AllFinish
     -- ^ The current test tree will be executed after its dependencies finish,
     -- regardless of whether they succeed or not.
-  deriving (Eq, Show)
+  deriving (Eq, Show, Read)
```
Bodigrim, line 253: Please add `@since` here and below where appropriate.

core/Test/Tasty/Core.hs
```diff
@@ -243,7 +250,21 @@ data DependencyType
...
+  -- ^ Execute tests one after another
+  | Parallel
+  -- ^ Execute tests in parallel
+  deriving (Show, Read)
+
+instance IsOption ExecutionMode where
+  defaultValue = Parallel
+  parseValue = readMaybe
+  optionName = Tagged "" -- internal option
+  optionHelp = Tagged "" -- internal option
```
Bodigrim, line 267: Did you check how `--help` looks now?

core/CHANGELOG.md
martijnbastiaan: That's correct, but _only_ for dependencies specified using `sequentialTestGroup`. Patterns defined using patterns retain their old behavior. (The initial approach did tackle this, but its implementation complexity was deemed too high.)

I see I didn't add a test for this, so I'll leave this conversation open until I've done that.

core/Test/Tasty/Core.hs
martijnbastiaan: Yeah, it doesn't print anything. But given that this is (now?) behavior we rely on I'll add a test.

martijnbastiaan: Ah wait, `IsOption` instances only get a corresponding command line flag if they implement `optionCLParser`. Otherwise they're internal anyway.

martijnbastiaan:
@Bodigrim All done :)

core/Test/Tasty/Core.hs
```diff
@@ -398,6 +432,17 @@ trivialFold = TreeFold
...
 
+
+-- | Indicates whether a test matched in an evaluated subtree. If no filter was
+-- used, tests always match.
+type TestMatched = Any
+
+-- | Used to force tests to be included, even if they would be filtered out by
+-- a user's filter. This is used to force dependencies of a test to run. For
+-- example, if test @A@ depends on test @B@ and test @A@ is selected to run, test
+-- @B@ will be forced to match.
```
Bodigrim, line 443: Does this apply universally or only to sequential test groups?

Bodigrim (commented):
@martijnbastiaan  I'm truly sorry that it takes so long, I'm very much overloaded recently. This is hopefully the last round, unless @andreasabel wants to take a look.

core/Test/Tasty/Core.hs
martijnbastiaan: Only sequential test groups, I've extended the documentation to highlight this.

core/Test/Tasty/Core.hs
```diff
@@ -478,23 +523,62 @@ evaluateOptions opts = \case
...
+
+-- | Filter test tree by pattern, replacing leafs with 'AnnEmptyTestTree'.
+filterByPattern :: AnnTestTree (OptionSet, Path) -> AnnTestTree OptionSet
+filterByPattern = snd . go mempty
+  where
+    go 
+      :: ForceTestMatch
+      -> AnnTestTree (OptionSet, Path)
+      -> (TestMatched, AnnTestTree OptionSet)
+    go forceMatch = \case
```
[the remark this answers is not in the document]
martijnbastiaan: It's on line 575 on the `Sequential` arm of the case:

```haskell
          Sequential _ ->
            second
              (AnnTestGroup opts name)
              (mapAccumR go forceMatch trees)
```

`mapAccumR` accumulates the "force match" value from right to left.

This is tested in `core-tests/SequentialTestGroup.hs`. The basic idea there is to build a test tree, and make each `TestTree` leaf insert its test ID into an `MVar [Int]`. At the very end, that list is checked for consistency. That is, it checks that tests that should have run first appear in the list first, etc.

core/Test/Tasty/Run.hs
```diff
@@ -382,25 +429,35 @@ createTestActions opts0 tree = do
...
+    goSeqGroup depType prevDeps treeM = do
+      tree0 <- local (second (prevDeps ><)) treeM
+
+      let
+        toDep TestAction {..} = Dependency depType (ExactDep testPath testStatus)
+        deps0 = Seq.fromList (toDep <$> collectTests tree0)
+
+        -- If this test tree is empty (either due to it being actually empty, or due
+        -- to all tests being filtered) we need to propagate the previous dependencies.
+        deps1 = if Seq.null deps0 then prevDeps else deps0
```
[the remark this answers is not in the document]
martijnbastiaan: This edge case was actually found by the tests :), see `core-tests/SequentialTestGroup.hs`. I've added an explicit test for this case, though it is also covered by the fuzz/quicktest checks.

core/Test/Tasty/Core.hs
martijnbastiaan: I've added `mkInternalOptionCLParser` to make this much more explicit. Is this what you were after?

martijnbastiaan:
> @martijnbastiaan I'm truly sorry that it takes so long, I'm very much overloaded recently. 

No worries, I appreciate the feedback. If it would be appreciated: I could help out reviewing PRs for this project and doing some general maintenance.

core/Test/Tasty/Core.hs
Bodigrim: `Test.Tasty.Options` is an exposed module, and I don't see much point in offering `mkInternalOptionCLParser` to users.

I meant literally `optionCLParser = mkOptionCLParser` [`internal`](https://hackage.haskell.org/package/optparse-applicative-0.17.0.0/docs/Options-Applicative.html#v:internal).

martijnbastiaan: Right, I wasn't aware of `internal`. Applied!

Bodigrim (approved):
LGTM, thanks! Unless there is any feedback from @andreasabel by the end of May, I'll proceed with a merge.

Bodigrim:
Thanks for persistence, @martijnbastiaan :)

`tasty` is a fairly stable package, so before we cut a new major release, I'd like to resolve as many other issues as possible. #341 and #311 are probably closest to completion, and #231 and #342 seem actionable too.

Bodigrim:
> If it would be appreciated: I could help out reviewing PRs for this project and doing some general maintenance.

@martijnbastiaan much appreciated! Sorry, I don't have admin rights to grant you formal privileges (and I'm reluctant to distract @UnkindPartition by asking), but you are most welcome.

martijnbastiaan:
I read about his situation, best not to distract him yeah. I'll see what I can do for the listed issues/PRs.

PR #344, opened 2022-07-31
Do not depend on unbounded-delays on 64-bit architecture

Bodigrim:
On 64-bit arch there is no point to depend on `unbounded-delays`: `(maxBound :: Int)` microseconds are almost 300 000 years. Not a terribly big deal, but shaving off a dependency is always nice. Basically, CI jobs for all packages, which use `tasty`, have one less thing to compile.

### DIFF
--- a/core/Test/Tasty/Run.hs
+++ b/core/Test/Tasty/Run.hs
@@ -11,6 +11,7 @@ module Test.Tasty.Run
 import qualified Data.IntMap as IntMap
 import qualified Data.Sequence as Seq
 import qualified Data.Foldable as F
+import Data.Int (Int64)
 import Data.Maybe
 import Data.Graph (SCC(..), stronglyConnComp)
 import Data.Typeable
@@ -21,7 +22,6 @@ import Control.Monad.Trans.Reader (ReaderT(..), local, ask)
 import Control.Monad.Trans.Writer (WriterT(..), execWriterT, mapWriterT, tell)
 import Control.Concurrent
 import Control.Concurrent.STM
-import Control.Concurrent.Timeout (timeout)
 import Control.Concurrent.Async
 import Control.Exception as E
 import Control.Applicative
@@ -29,6 +29,12 @@ import Control.Arrow
 import Data.Monoid (First(..))
 import GHC.Conc (labelThread)
 import Prelude  -- Silence AMP and FTP import warnings
+
+#ifdef MIN_VERSION_unbounded_delays
+import Control.Concurrent.Timeout (timeout)
+#else
+import System.Timeout (timeout)
+#endif
 
 import Test.Tasty.Core
 import Test.Tasty.Parallel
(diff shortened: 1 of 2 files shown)


Bodigrim:
@VictorCMiraldo @andreasabel this is ready for review.

core/Test/Tasty/Run.hs
```diff
@@ -171,7 +177,9 @@ executeTest action statusVar timeoutOpt inits fins = mask $ \restore -> do
             , resultTime = fromIntegral t
             , resultDetailsPrinter = noResultDetails
             }
-      fromMaybe timeoutResult <$> timeout t a
+      -- If compiled with unbounded-delays then t' :: Integer, otherwise t' :: Int
+      let t' = fromInteger (min (max 0 t) (toInteger (maxBound :: Int64)))
```
VictorCMiraldo, line 181: Is the `max 0 t` necessary? Both [`System.Timeout.timeout`](https://hackage.haskell.org/package/base-4.8.2.0/docs/src/System.Timeout.html#timeout) and its [`unbounded-delays` counterpart](https://hackage.haskell.org/package/unbounded-delays-0.1.1.1/docs/src/Control.Concurrent.Timeout.html#timeoutWithPred) are `fmap Just` when the specified delay is less than zero.

That being said, it doesn't hurt to keep it there. :)

VictorCMiraldo (approved):
Pardon my delay, I just came back from holidays. 
Shaving off a dependency is always nice and the changes are really small, thanks @Bodigrim!

core/Test/Tasty/Run.hs
Bodigrim: `max 0 t` is to prevent underflow. Imagine `t = -(2^63+1) :: Integer`: previously it meant no delay, but without `max 0 t` it would be cast to `2^63-1 :: Int`, a huge timeout.

core/tasty.cabal
```diff
@@ -70,8 +70,13 @@ library
     transformers         >= 0.5,
     tagged               >= 0.5,
     optparse-applicative >= 0.14,
-    unbounded-delays     >= 0.1,
     ansi-terminal        >= 0.9
+
+  -- No reason to depend on unbounded-delays on 64-bit architecture
+  if(!arch(x86_64) && !arch(aarch64))
```
andreasabel, line 76: This looks brittle.
What are the possible values of `arch`?  I couldn't even find the list in the cabal docs...
At least, checking it this way (with a whitelist) is conservative.   But new architectures might easily fall under the radar.

Bodigrim: Full list is here: https://hackage.haskell.org/package/Cabal-3.8.1.0/docs/src/Distribution.Simple.PreProcess.html#local-6989586621679281986

Potentially `powerpc64`, `ia64` and `s390x` are also guaranteed to be 64-bit, but I decided to whitelist only architectures I have access to.

andreasabel: I suggest to open an issue to potentially extend the list of recognized architectures, and whoever has access to one of them and can do testing can attack it.

Issue #288, opened 2021-01-02
Cannot use tasty with GHCJS (at least clock is problematic)


phadej:
```
uncaught exception in Haskell main thread: ReferenceError: h$clock_gettime is not defined
ReferenceError: h$clock_gettime is not defined
```

feuerbach:
I pushed a potential fix to the `fix-ghcjs` branch, could you try it?

phadej:
That seems to work!

feuerbach:
Cool, released as tasty-1.4.0.2.

phadej:
Thanks!

PR #345, opened 2022-08-08
Do not depend on clock package, decomission clock flag

Bodigrim:
Starting from GHC 8.4 we can use `GHC.Clock` from `base`, and for diminishing minority of users of GHC < 8.4 `time` is sufficiently good (and does not require build, because it is a boot package).

### DIFF
--- a/core/Test/Tasty/Runners/Utils.hs
+++ b/core/Test/Tasty/Runners/Utils.hs
@@ -7,15 +7,15 @@ import Control.Exception
 import Control.Applicative
 import Control.Concurrent (mkWeakThreadId, myThreadId)
 import Control.Monad (forM_)
-#ifndef VERSION_clock
-import Data.Time.Clock.POSIX (getPOSIXTime)
-#endif
 import Data.Typeable (Typeable)
 import Prelude  -- Silence AMP import warnings
 import Text.Printf
 import Foreign.C (CInt)
-#ifdef VERSION_clock
-import qualified System.Clock as Clock
+
+#if MIN_VERSION_base(4,11,0)
+import GHC.Clock (getMonotonicTime)
+#else
+import Data.Time.Clock.POSIX (getPOSIXTime)
 #endif
 
 -- Install handlers only on UNIX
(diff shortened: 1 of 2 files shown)


VictorCMiraldo (approved):
:tada: less dependencies is always nice! LGTM!

andreasabel (approved):

andreasabel:
> for diminishing minority of users of GHC < 8.4 `time` is sufficiently good

I take your word for it...

Bodigrim:
@andreasabel the difference between `clock` and `time` is observable only if a user changes system clock while running tests. System time change can lead to spurious timeouts, if `time` is used. But I seriously doubt plausibility of such scenario: nowadays builds with GHC < 8.4 are mostly happening on CI, not on user machines. And any timeout can fail depending on current system activity anyway. 

Note that `clock` flag, switching between `clock` and `time`, was introduced for GHCJS, which does not support FFI calls required for `clock`.

ncfavier:
Reintroduces https://github.com/UnkindPartition/tasty/issues/288 in a different form: `ReferenceError: h$getMonotonicNSec is not defined`. This was apparently known so I'm not sure why this was merged?

Bodigrim:
What exactly was “apparently known”?

No CI job = not supported. If you want us to consider GHCJS support, please contribute one.

andreasabel:
If there is a macro akin to [`__GLASGOW_HASKELL__`](https://downloads.haskell.org/ghc/latest/docs/users_guide/phases.html) in GHCJS, once could add it to the conditions of `import GHC.Clock`.
https://github.com/UnkindPartition/tasty/blob/9ff66c2339ba72711185111afe9eddbf3ce5e0fc/core/Test/Tasty/Runners/Utils.hs#L15-L19

ncfavier:
There is `ghcjs_HOST_OS`, but I think that would also affect the new GHC 9.6 JS backend (which is not to be confused with GHCJS, although it uses `ghcjs` as its OS name...), which seems to support GHC.Clock.

Anyway, since GHCJS is on the way out anyway I've just [patched](https://github.com/NixOS/nixpkgs/pull/230673/files#diff-23ecfe4f32ee3203aa1fe2663504d2294c0e4bb69836df5de72003dec50ea62dR120-R129) this in nixpkgs.

PR #347, opened 2022-08-10
Add GHC 9.4 CI job

Bodigrim:


.github/workflows/ci.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest]
-        ghc: ['8.0', '8.2', '8.4', '8.6', '8.8', '8.10', '9.0', '9.2']
+        ghc: ['8.0', '8.2', '8.4', '8.6', '8.8', '8.10', '9.0', '9.2', '9.4.1']
```
andreasabel, line 17: Since you are departing from the _x.y_ scheme for 9.4.1, please add a comment there why you do so.

andreasabel (changes requested):

.github/workflows/ci.yaml
Bodigrim: Done.

andreasabel (approved):
Thanks!

Issue #350, opened 2022-09-16
Assertion failures don't show full stacktrace


smatting:
It can be hard to figure out where in a test case assertions are failing. For example if the test case has multiple call sites of a helper function that might fail it's impossible to figure out which call site is raising the HUnitFailure, because only the lower most SrcLoc of the callstack is saved.

I propose to always include the full stack trace in HUniFailure.

Without a full stackrace you can be lost:
```
Tests
  Auth
    LoginId parser: valid email: FAIL
      test/unit/Test/Wire/API/User/Auth.hs:34:
      Some assertion failed. Try to find me!
```

In contrast:
```
Tests
  Auth
    LoginId parser: valid email: FAIL
      Error message: Some assertion failed. Try to find me!
      
      CallStack (from HasCallStack):
        assertFailure, called at test/unit/Test/Wire/API/User/Auth.hs:52:11 in main:Test.Wire.API.User.Auth
        helper2, called at test/unit/Test/Wire/API/User/Auth.hs:49:10 in main:Test.Wire.API.User.Auth
        helper, called at test/unit/Test/Wire/API/User/Auth.hs:34:3 in main:Test.Wire.API.User.Auth
```

PR #351, opened 2022-09-16
Show full stacktrace in assertion failures

smatting:
Implements this proposal: https://github.com/UnkindPartition/tasty/issues/350

### DIFF
--- a/hunit/Test/Tasty/HUnit/Orig.hs
+++ b/hunit/Test/Tasty/HUnit/Orig.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DeriveDataTypeable, FlexibleInstances, TypeSynonymInstances #-}
+{-# LANGUAGE DeriveDataTypeable, FlexibleInstances, TypeSynonymInstances, RecordWildCards #-}
 
 -- required for HasCallStack by different versions of GHC
 {-# LANGUAGE ConstraintKinds, FlexibleContexts #-}
@@ -11,6 +11,7 @@ import qualified Control.Exception as E
 import Control.Monad
 import Data.Typeable (Typeable)
 import Data.CallStack
+import Data.List
 
 -- Interfaces
 -- ----------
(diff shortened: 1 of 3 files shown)


Bodigrim:
What’s the behavior of `hunit` proper? Does it show full stack traces?

smatting:
I don't think so. I was looking [at this](https://hackage.haskell.org/package/HUnit-1.6.2.0/docs/src/Test.HUnit.Base.html#performTest).

hunit/Test/Tasty/HUnit/Orig.hs
```diff
@@ -133,15 +129,33 @@ instance (AssertionPredicable t) => AssertionPredicable (IO t)
...
+prettyCallStack :: CallStack -> String
+prettyCallStack = intercalate "\n" . prettyCallStackLines
+
+prettyCallStackLines :: CallStack -> [String]
+prettyCallStackLines cs = case cs of
+  []  -> []
+  stk -> "CallStack (from HasCallStack):"
+       : map (("  " ++) . prettyCallSite) stk
+  where
+    prettyCallSite (f, loc) = f ++ ", called at " ++ prettySrcLoc loc
```
martijnbastiaan, line 149: `base` re-exports `prettyCallStackLines`, so we should be able to use that. I'm assuming you made a local definition because you didn't want to rely on the `ghc` package?

martijnbastiaan (commented):
This seems like a sensible change to me, I've been very slightly annoyed by one-line stack traces in Tasty output before. I do think it would warrant a changelog entry.

martijnbastiaan:
I see. @smatting Would you mind opening a PR against `hunit` to do the same?

MangoIV:
Do I understand correctly, that to make this progress, we would first have to implement this proposal on the HUnit side of things?

Bodigrim:
@MangoIV not necessarily, it is blocked on #388 in the first place.

MangoIV:
aha, makes sense, do you perhaps happen to have insight on the decision process there?

hunit/Test/Tasty/HUnit/Orig.hs
MangoIV: this actually has a different reason, namely that the `CallStack` here is not the same that is used in `GHC.Exception`.

Bodigrim:
So far no one volunteered "to own `tasty-hunit` and make decisions about its future", this would be the first step to make any progress on #388.

PR #362, opened 2023-01-10
Add since pragmas

Bodigrim:


andreasabel (approved):
Wow that must have involved quite some digging.
Thanks!
(I am not going to duplicate your work and check all the `@since` annotations for correctness.)

PR #363, opened 2023-03-09
Split foldTestTree

Bodigrim:
https://github.com/UnkindPartition/tasty/blob/0debac85701560e8c96cd3705988c50197cb214e/core/Test/Tasty/Core.hs#L410-L427

At the moment `foldTestTree` does three jobs at once: evaluates options, filters by pattern and actually folds. This PR splits it into three functions, each with its own task. 

This refactoring should hopefully guide a better way for #343, but separating concerns is a good thing anyways.

### DIFF
--- a/core/Test/Tasty/Core.hs
+++ b/core/Test/Tasty/Core.hs
@@ -393,10 +393,6 @@ trivialFold = TreeFold
 --
 -- Thus, it is preferred to an explicit recursive traversal of the tree.
 --
--- Note: right now, the patterns are looked up only once, and won't be
--- affected by the subsequent option changes. This shouldn't be a problem
--- in practice; OTOH, this behaviour may be changed later.
---
 -- @since 0.7
 foldTestTree
   :: forall b . Monoid b
--- a/core/Test/Tasty/Core.hs
+++ b/core/Test/Tasty/Core.hs
@@ -1,6 +1,7 @@
 -- | Core types and definitions
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
 module Test.Tasty.Core
   ( FailureReason(..)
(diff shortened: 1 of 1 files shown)


core/Test/Tasty/Core.hs
```diff
@@ -408,23 +405,57 @@ foldTestTree
...
+  where
+    go :: TestTreeWithOptions -> b
+    go = \case
+      EmptyTestTreeWO               -> mempty
+      SingleTestWO opts name test   -> fTest opts name test
+      TestGroupWO opts name trees   -> fGroup opts name (foldMap go trees)
+      WithResourceWO opts res0 tree -> fResource opts res0 $ \res -> go (tree res)
+      AfterWO opts deptype dep tree -> fAfter opts deptype dep (go tree)
+
+-- | 'TreeFold' is an eliminator for 'TestTreeWithOptions'.
```
martijnbastiaan, line 418: I'm wasn't familiar with the concept `eliminator`. Googling `eliminator haskell` gave 7 results pointing to the Hackage package `eliminators`, which doesn't go through any effort trying to explain what the concept is. More looking around only connects this term to dependent types which I don't think directly applies here.

In short; I think this comment needs to be reworded, preferably without use of very specific jargon.

martijnbastiaan (commented):
While I don't think this is going to help for #343, I agree this code is cleaner. I feel it would be inappropriate to flag it approve/request changes given that I'm not a maintainer, so I'm going to just mark it as a comment :).

--------

Fundamentally I think the problem over at #343 is that in `foldResource`:

```haskell
  , foldResource :: forall a . OptionSet -> ResourceSpec a -> (IO a -> b) -> b
```

there is no way (?) to bubble up test match information as it can't "escape" the lambda:

```haskell
      WithResourceWO opts res0 tree ->
        WithResourceWO opts res0 $ \res -> go path (tree res)
```

Bodigrim:
@martijnbastiaan I did not want to overengineer it from the get-go, but the idea is that you can out arbitrary annotations instead of `OptionSet`. I've updated the PR.

martijnbastiaan:
Looks good to me!

core/Test/Tasty/Core.hs
```diff
@@ -408,23 +405,57 @@ foldTestTree
...
+      AnnWithResource opts res0 tree -> fResource opts res0 $ \res -> go (tree res)
+      AnnAfter opts deptype dep tree -> fAfter opts deptype dep (go tree)
+
+-- | 'TestTree' with arbitrary annotations, e. g., evaluated 'OptionSet'.
+data AnnTestTree ann
+  = AnnEmptyTestTree
+  | forall t . IsTest t => AnnSingleTest ann TestName t
+  | AnnTestGroup ann TestName [AnnTestTree ann]
+  | forall a . AnnWithResource ann (ResourceSpec a) (IO a -> AnnTestTree ann)
+  | AnnAfter ann DependencyType Expr (AnnTestTree ann)
```
andreasabel, line 424: Please give a haddock comment for each of the constructors.

If the original `TestTree` has such comments, a minimal doc here would be to link to these original constructors so I can get the explanation by a single click.  But more doc is always welcome!

andreasabel (changes requested):
@Bodigrim : I only did a formal review since I am not very familiar with the matter.

In general, seems like a good idea to factorize the fold.

core/Test/Tasty/Core.hs
Bodigrim: Done, but this data type is not exported any way.

Bodigrim:
@martijnbastiaan does it look closer to what could be helpful for your PR?

martijnbastiaan:
Looks good to me! This is going to make it a bit easier to reason about the changes we'd need to make over there.

andreasabel (approved):
Thanks!

martijnbastiaan:
I'm rebasing my PR on this one to see if there are any unexpected things in here.

PR #367, opened 2023-04-28
resolve #298: make color formats configurable via environment variables

### LINKED ISSUE #298: Font color in some contexts might be indistinguishable from terminal foreground color
I noticed this when running tests and benchmarks of `bytestring`, which uses tasty (and tasty-bench). Compare
Screenshot_20210419_013308
vs
Screenshot_20210419_013334

So far this looks like a tasty's issue, as I don't think I ever noticed anything that might hint it's a color scheme issue.

bfrk:


Bodigrim:
Closing in favor of #369. Thanks for you contribution anyways!

PR #369, opened 2023-05-18
Fix colors in console reporter, resolves #298

### LINKED ISSUE #298: Font color in some contexts might be indistinguishable from terminal foreground color

Bodigrim:
No need to be clever, just `putStr` :)

### DIFF
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -709,7 +709,9 @@ ok, fail, skipped, infoOk, infoFail :: (?colors :: Bool) => String -> IO ()
 fail     = output failFormat
 ok       = output okFormat
 skipped  = output skippedFormat
-infoOk   = output infoOkFormat
+-- Just default foreground color for 'infoOk'; do not apply 'infoOkFormat',
+-- because terminal's background could be white itself. See #298.
+infoOk   = putStr
 infoFail = output infoFailFormat
 
 output


core/Test/Tasty/Ingredients/ConsoleReporter.hs
```diff
@@ -709,7 +709,9 @@ ok, fail, skipped, infoOk, infoFail :: (?colors :: Bool) => String -> IO ()
 fail     = output failFormat
 ok       = output okFormat
 skipped  = output skippedFormat
-infoOk   = output infoOkFormat
+-- Just default foreground color for 'infoOk'; do not apply 'infoOkFormat',
+-- because terminal's background could be white itself. See #298.
```
andreasabel, line 713: Yeah, that is the usual problem with coloring output.

We have the same problem when considering coloring errors for Agda.
- https://github.com/agda/agda/pull/6544

It would be great to have _symbolic_ colors, that resolve to physical color according to some (standard or custom) coloring scheme, depending on the background color.
I haven't found any Haskell solution to that problem yet (that at least works well on modern terminals).

Bodigrim: Yeah, I'm not aware of any fundamental solution. Other colors used by `tasty` are less problematic: it's highly unlikely that someone has a terminal with red or green background; it's only `infoOkFormat` who poses an issue.

andreasabel (approved):
Of course I approve.

PR #370, opened 2023-06-02
CI: disable haddocks for GHC 8.2

Bodigrim:
The last release of `optparse-applicative` brought `text` into dependencies, and now it chokes Haddock with GHC 8.2: https://github.com/UnkindPartition/tasty/actions/runs/5147520656/jobs/9268225724. Cf. https://github.com/haskell/text/pull/478. 

I honestly see little value in indulging obsolete versions of Haddock, so I suggest just removing this build step from CI job for GHC 8.2.


andreasabel (approved):
Thanks!

PR #371, opened 2023-06-22
Don't depend on unbounded-delays for riscv64

felixonmars:
64-bit RISC-V should fall into the same case in
https://github.com/UnkindPartition/tasty/pull/344 and should be added here. Tested locally.

### DIFF
--- a/core/tasty.cabal
+++ b/core/tasty.cabal
@@ -68,7 +68,7 @@ library
     ansi-terminal        >= 0.9  && < 1.2
 
   -- No reason to depend on unbounded-delays on 64-bit architecture
-  if(!arch(x86_64) && !arch(aarch64) && !arch(ppc64) && !arch(s390x))
+  if(!arch(x86_64) && !arch(aarch64) && !arch(ppc64) && !arch(s390x) && !arch(riscv64))
     build-depends:
       unbounded-delays >= 0.1 && < 0.2
 


felixonmars:
I actually thought I did... Now opened as https://github.com/haskell/cabal/pull/9062

amesgen:
> @amesgen @TerrorJack any ideas what's up with our WASM CI job?

Ah sorry, this is due to the respective bindist from the GHC CI expiring. I have an [upstream MR](https://gitlab.haskell.org/ghc/ghc-wasm-meta/-/merge_requests/29) to fix this properly in ghc-wasm-meta. This should be merged soon-ish; will create a PR as soon as it. To fix this immediately, one can bump
https://github.com/UnkindPartition/tasty/blob/b509e0470fc27ba64a18c185611eee5ef805a7c7/.github/workflows/ci.yaml#L86
to [bd9533e34df53694a4c7e4102fced1fdc0596024](https://gitlab.haskell.org/ghc/ghc-wasm-meta/-/commit/bd9533e34df53694a4c7e4102fced1fdc0596024), do you want to do this here or shall I create a PR?

Bodigrim:
Unfortunately, this would have to wait not only until Cabal 3.12 release itself (which is soon), but until Hackage upgrades to it.

Bodigrim:
Thanks, we can finally merge this now!

PR #376, opened 2023-07-15
Propagate 'forceMatch' for empty test groups

martijnbastiaan:
Similar to the logic in `goSeqGroup`:

https://github.com/UnkindPartition/tasty/blob/804b2a7d8323902ece3f1077c36860929343d9db/core/Test/Tasty/Run.hs#L442-L444

`filterByPattern` needs to return `forceMatch`, lest the return value will always be `Any False` due to `(mempty :: Any) ~ Any False` for the `Parallel` branch:

https://github.com/UnkindPartition/tasty/blob/804b2a7d8323902ece3f1077c36860929343d9db/core/Test/Tasty/Core.hs#L568-L572

Note that this only applies to `testGroup` (i.e. `Parallel`), but fixing it in that specific branch would create yet another level of nesting in `filterByPattern`.

An empty test group has been added to the `filterTestTree` test logic in `SequentialTestGroup` to prevent regressions.

### DIFF
--- a/core-tests/SequentialTestGroup.hs
+++ b/core-tests/SequentialTestGroup.hs
@@ -89,6 +89,7 @@ filterTestTree pattern =
       , sequentialTestGroup "E" AllSucceed
         [ emptyTest "F"
         , emptyTest "G"
+        , testGroup "XX" []
         , emptyTest "H"
         ]
     ]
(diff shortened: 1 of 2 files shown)


martijnbastiaan:
Thank you :)

PR #380, opened 2023-08-02
Allow --pattern|-p option to repeat

rhendric:
The effect of multiple -p options is the same as &&-ing the expressions together, but allowing them to be specified in separate options makes scripting test commands simpler.

---

Motivation: I want to run tests with a small shell script such as the following:

```
#!/usr/bin/env bash
default_test_options=(...)
test-exe "${default_test_options[@]}" "$@"
```

If `default_test_options` contains a pattern filter like `-p ! /very-slow-test/`, then I'm currently unable to provide any options to the script to further focus on a set of interest. I would have to write some extra logic to the script to detect pattern filters and inject them into the default pattern filter, making the script much more complicated than I would like it to be.

Allowing multiple `-p` options to be provided to Tasty programs resolves the issue.

### DIFF
--- a/core-tests/core-tests.cabal
+++ b/core-tests/core-tests.cabal
@@ -50,3 +50,9 @@ executable failing-pattern-test
   build-depends:       base <= 5, tasty, tasty-hunit, random >= 1.2, mtl
   default-extensions:  ScopedTypeVariables
   ghc-options:         -Wall -fno-warn-type-defaults
+
+executable multiple-pattern-test
+  import:              commons
+  main-is:             multiple-pattern-test.hs
+  build-depends:       base < 5, tasty, tasty-hunit
+  ghc-options:         -Wall -fno-warn-type-defaults
--- a/core/Test/Tasty/Patterns.hs
+++ b/core/Test/Tasty/Patterns.hs
@@ -1,6 +1,6 @@
 -- | Test patterns
 
-{-# LANGUAGE CPP, DeriveDataTypeable #-}
+{-# LANGUAGE CPP, DeriveDataTypeable, TypeApplications #-}
 
 module Test.Tasty.Patterns
   ( TestPattern(..)
@@ -18,6 +18,9 @@ import Test.Tasty.Patterns.Parser
 import Test.Tasty.Patterns.Eval
 
 import Data.Char
+import Data.Coerce (coerce)
+import Data.List.NonEmpty (nonEmpty)
+import Data.Maybe (catMaybes)
 import Data.Typeable
 import Options.Applicative hiding (Success)
 #if !MIN_VERSION_base(4,11,0)
@@ -39,12 +42,16 @@ newtype TestPattern =
 noPattern :: TestPattern
 noPattern = TestPattern Nothing
 
+-- | Since tasty-1.5, this option can be specified multiple times on the
+-- command line. Only the tests matching all given patterns will be selected.
 instance IsOption TestPattern where
   defaultValue = noPattern
   parseValue = parseTestPattern
   optionName = return "pattern"
   optionHelp = return "Select only tests which satisfy a pattern or awk expression"
-  optionCLParser = mkOptionCLParser (short 'p' <> metavar "PATTERN")
+  optionCLParser =
+    fmap (TestPattern . fmap (foldr1 And) . nonEmpty . catMaybes . coerce @[TestPattern]) . some $
+      mkOptionCLParser (short 'p' <> metavar "PATTERN")
 
 -- | @since 1.2
 parseExpr :: String -> Maybe Expr
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -69,6 +69,7 @@ jobs:
         core-tests/exit-status-tests.sh
         # Prevent Git for Windows from replacing slashes with backslashes in patterns
         MSYS_NO_PATHCONV=1 core-tests/failing-pattern-test.sh
+        core-tests/multiple-pattern-test.sh
 
     - name: Test resource-release-test.sh
       if: runner.os != 'Windows'
--- a/core-tests/multiple-pattern-test.hs
+++ b/core-tests/multiple-pattern-test.hs
@@ -1,0 +1,14 @@
+import Test.Tasty
+import Test.Tasty.HUnit
+
+main :: IO ()
+main = defaultMain $ testGroup "all"
+  [ testGroup "red"
+    [ testCase "square" $ pure ()
+    , testCase "circle" $ pure ()
+    ]
+  , testGroup "green"
+    [ testCase "square" $ pure ()
+    , testCase "circle" $ pure ()
+    ]
+  ]
--- a/core-tests/multiple-pattern-test.sh
+++ b/core-tests/multiple-pattern-test.sh
@@ -1,0 +1,28 @@
+#!/bin/sh
+
+set -eux
+
+if ! command -v multiple-pattern-test
+then
+  echo "multiple-pattern-test executable is not in PATH, aborting"
+  exit 1
+fi
+
+[ "$(multiple-pattern-test -l | wc -l)" -eq 4 ]
+[ "$(multiple-pattern-test -l -p red | wc -l)" -eq 2 ]
+[ "$(multiple-pattern-test -l -p circle | wc -l)" -eq 2 ]
+[ "$(multiple-pattern-test -l -p red -p circle | wc -l)" -eq 1 ]
+[ "$(multiple-pattern-test -l -p red -p circle -p green | wc -l)" -eq 0 ]
+
+# Edge case: the empty pattern matches everything
+[ "$(multiple-pattern-test -l -p '' | wc -l)" -eq 4 ]
+[ "$(multiple-pattern-test -l -p '' -p '' | wc -l)" -eq 4 ]
+[ "$(multiple-pattern-test -l -p '' -p red | wc -l)" -eq 2 ]
+[ "$(multiple-pattern-test -l -p red -p '' | wc -l)" -eq 2 ]
+[ "$(multiple-pattern-test -l -p '' -p red -p '' | wc -l)" -eq 2 ]
+[ "$(multiple-pattern-test -l -p red -p '' -p circle | wc -l)" -eq 1 ]
+
+# Environment variable is entirely overridden by any command line options
+[ "$(TASTY_PATTERN=red.circle multiple-pattern-test -l | wc -l)" -eq 1 ]
+[ "$(TASTY_PATTERN=red.circle multiple-pattern-test -l -p square | wc -l)" -eq 2 ]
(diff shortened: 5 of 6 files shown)


core/Test/Tasty/Patterns.hs
```diff
@@ -28,23 +29,27 @@ import Data.Monoid
 newtype TestPattern =
   -- | @since 1.1
   TestPattern
-    (Maybe Expr)
+    [Expr]
```
Bodigrim, line 32: I like the idea, but I don't think you need to change the definition of `TestPattern`, this is disruptive. Instead you should be able to use `foldl' Test.Tasty.Patterns.Types.And` immediately in `optionCLParser`, combining everything into a single `Expr`.

rhendric: Good idea, that's a much smaller change.

Bodigrim (approved):

core/Test/Tasty/Patterns.hs
```diff
@@ -44,7 +45,11 @@ instance IsOption TestPattern where
   parseValue = parseTestPattern
   optionName = return "pattern"
   optionHelp = return "Select only tests which satisfy a pattern or awk expression"
-  optionCLParser = mkOptionCLParser (short 'p' <> metavar "PATTERN")
+  optionCLParser = fmap (foldl1' testPatternAnd) $ some $ mkOptionCLParser (short 'p' <> metavar "PATTERN")
```
Bodigrim, line 48: Just to be on the safe side and avoid a partial function, how about `foldl' testPatternAnd noProgress` instead of `foldl1' testPatternAnd`? I understand that `some` returns a non-empty list, but this unfortunately not reflected in types.

Bodigrim:
CC @martijnbastiaan for review

martijnbastiaan (commented):
Cool, I've wanted this myself before! Two things:

* I do think this deserves an entry in `core/CHANGELOG.md`
* If it's not too much trouble, I think one or two simple tests should be added to `core-tests`

andreasabel (changes requested):
My intuitive understanding of passing `-p` several times would be that any test matching any of the patterns is run.  This would be _disjunction_.  If I am not mistaken, this PR implements the _conjunction_, which I find unintuitive.
Since there are two monoids (conjunction and disjunction), and people have different ideas which one should be the default, I think we should not arbitrarily decide on either.

Making an option monoidal also violates the invariant that each option can equivalently be given by an environment variable.  But you cannot have monoidal env vars. https://github.com/UnkindPartition/tasty/blob/959fe9164c7bda395799040242726374aeeff9d6/README.md?plain=1#L245-L247

So, unfortunately, my judgement is "reject".

rhendric:
> Since there are two monoids (conjunction and disjunction), and people have different ideas which one should be the default, I think we should not arbitrarily decide on either.

But it isn't arbitrary. Of the two monoids, only conjuction has a mempty consistent with what Tasty does when no `-p` options are provided: run all the tests.

> But you cannot have monoidal env vars.

I don't understand this objection. Sure, the interface of environment variables doesn't let you specify a variable multiple times. But it's still the case that anything you can specify with command line options can also be specified with env vars. It's also still the case that environment variables are overridden if any matching options are provided on the command line. What problems would this cause?

core/Test/Tasty/Patterns.hs
andreasabel: I think that `some` (as opposed to `many`) returns a non-empty list is an essential invariant one should rely on _and_ reflect this in its use.  So `foldl1'` is the right choice imho.
Not exploiting the invariants blurs the picture imo and leads to less readable code.

core/Test/Tasty/Patterns.hs
andreasabel: Mmh, somehow I liked the `[Expr]` better here (with a comment that it is to be understood conjunctively) because this better reflects what is going on here.
Of course, this would be an API change then and require a major version bump.

core/Test/Tasty/Patterns.hs
```diff
@@ -44,7 +45,11 @@ instance IsOption TestPattern where
   parseValue = parseTestPattern
   optionName = return "pattern"
   optionHelp = return "Select only tests which satisfy a pattern or awk expression"
-  optionCLParser = mkOptionCLParser (short 'p' <> metavar "PATTERN")
+  optionCLParser = fmap (foldl' testPatternAnd noPattern) $ some $ mkOptionCLParser (short 'p' <> metavar "PATTERN")
+    where
+    testPatternAnd (TestPattern Nothing) (TestPattern mbY) = TestPattern mbY
+    testPatternAnd (TestPattern mbX) (TestPattern Nothing) = TestPattern mbX
+    testPatternAnd (TestPattern (Just x)) (TestPattern (Just y)) = TestPattern (Just (And x y))
```
andreasabel, line 52: It feels that the `Nothing` cases should be impossible.
However, there is the weird behavior `parseTestPattern "" = Just noPattern`.
https://github.com/UnkindPartition/tasty/blob/c2cefb1839ead2713cf40a67463a0e4a0ea24e08/core/Test/Tasty/Patterns.hs#L57-L60
I would have expected `parseTestPattern "" = Nothing` (or even `parseTestPattern "" = undefined`), because
> The option `-p` expects an argument.

I would argue that the impossible cases should be marked in the source code as such,
```haskell
testPatternAnd (TestPattern (Just x)) (TestPattern (Just y)) = TestPattern (Just (And x y))
testPatternAnd _ _ = undefined
```
and covered by tests.

andreasabel (commented):
I am fine with this feature in general, but it needs to be accompanied by tests and documentation updates.

rhendric:
How and where should tests be written for this? core-tests looks a little idiosyncratic; am I supposed to test this by creating a new executable in core-tests.cabal with accompanying .hs and .sh files? And then how are those things invoked?

core/Test/Tasty/Patterns.hs
Bodigrim: Ideally it should be [`Data.Foldable1.foldl1'`](https://hackage.haskell.org/package/base-4.18.0.0/docs/Data-Foldable1.html#v:foldl1-39-) and [`Options.Applicative.NonEmpty.some1`](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative-NonEmpty.html#v:some1). We cannot use the former, because it's too new (just added in `base-4.18`), but `some1` is probably a good choice to return a proper type.

@rhendric sorry for bikeshedding of this line :)

core/Test/Tasty/Patterns.hs
Bodigrim: Unfortunately, `[Expr]` triggers changes in many other functions throughout the package, which is way too much breakage IMO.

core/Test/Tasty/Patterns.hs
Bodigrim: `mkOptionCLParser` is too restrictive, but it should be possible to avoid concatenation of `Maybe Expr` entirely with something like 
```haskell
optionCLParser = optional $ fmap (foldl1' testPatternAnd) $ some $ bla-bla-bla (short 'p' <> metavar "PATTERN")
```
for a certain `bla-bla-bla` ;)

Bodigrim:
> am I supposed to test this by creating a new executable in core-tests.cabal with accompanying .hs and .sh files?

That would be fine from my perspective. 

> And then how are those things invoked?

Please check `.github/workflows/ci.yml`.

core/Test/Tasty/Patterns.hs
rhendric: I don't think that's possible, if you intend to preserve the `parseTestPattern "" = Just noPattern` behavior. The parse action can return a value or raise an error; if the string it receives is empty, it can't return a `Nothing` (or we're stuck catenating `Maybe`s), it can't return a `TestPattern` (because that wraps `Maybe`s), and it can't return an `Expr` (unless you want me to stuff the empty string into an `ERE` and then test it outside before `And`ing it to another `Expr`, which is just like catenating `Maybe`s but worse). What's left for it to return in that case?

core/Test/Tasty/Patterns.hs
rhendric: Not having `Foldable1.foldl1'`, which of the following should I use?
* `Data.List.foldl1'`: takes a list, not a `NonEmpty`; so either I can't use `some1`, or I have to convert it to a list in the middle
* `Data.Foldable.foldl1`: not strict, probably a terrible idea, I'm only mentioning it out of completeness
* `Data.Foldable.foldr1`: I'm inclined to use this if you want me to use `some1`, but since you suggested `foldl'` first I want to be sure we're on the same page

rhendric:
I've written tests, added a changelog entry, and taken another approach with `optionCLParser` that will probably make everyone angry.

Bodigrim:
> ...and taken another approach with `optionCLParser` that will probably make everyone angry.

I'm actually quite happy with it :) @andreasabel @martijnbastiaan opinions?

martijnbastiaan (approved):

Bodigrim:
@andreasabel just a gentle reminder to review, I'd like to wrap up things for a release soon.

core/Test/Tasty/Patterns.hs
```diff
@@ -44,7 +47,9 @@ instance IsOption TestPattern where
   parseValue = parseTestPattern
   optionName = return "pattern"
   optionHelp = return "Select only tests which satisfy a pattern or awk expression"
-  optionCLParser = mkOptionCLParser (short 'p' <> metavar "PATTERN")
+  optionCLParser =
```
Bodigrim, line 50: @rhendric could you please attach a comment to the `instance IsOption TestPattern` saying "Since tasty-1.5 this command-line option can be specified multiple times with accumulating effect bla-bla-bla" or similar?

rhendric: Done.

core-tests/core-tests.cabal
```diff
@@ -50,3 +50,9 @@ executable failing-pattern-test
   build-depends:       base <= 5, tasty, tasty-hunit, random >= 1.2, mtl
   default-extensions:  ScopedTypeVariables
   ghc-options:         -Wall -fno-warn-type-defaults
+
+executable multiple-pattern-test
+  import:              commons
+  main-is:             multiple-pattern-test.hs
+  build-depends:       base <= 5, tasty, tasty-hunit
```
andreasabel, line 57: PVP upper bounds need to be strict, so, `base < 5`.
Why? Because the current constraint allows `base-5`, but not e.g. `base-5.0.0.0`, which does not make any sense.  The intention is clearly to only allow `base-4.*`.
This seems to be a systematic error in this file, so blame goes to someone else.

andreasabel (approved):
Thanks for the test!
I tried `cabal run multiple-pattern-test` locally.

core-tests/core-tests.cabal
rhendric: Changed.

andreasabel: Thanks!

Bodigrim:
Thanks @rhendric!

PR #387, opened 2023-09-11
Teach the HUnit Steps provider to log progress

michaelpj:
A cheap improvement to also show the steps as the tests are running. This does mean that we need to bump the `tasty` lower bound to 1.5.

### DIFF
--- a/hunit/CHANGELOG.md
+++ b/hunit/CHANGELOG.md
@@ -1,6 +1,12 @@
 Changes
 =======
 
+Unreleased
+--------------
+
+* Teach `testCaseSteps` to log progress
+  ([#387](https://github.com/UnkindPartition/tasty/pull/387)).
+
 Version 0.10.1
 ---------------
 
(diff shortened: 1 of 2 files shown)


Bodigrim:
> This does mean that we need to bump the `tasty` lower bound to 1.5.

Not necessarily: IINM the code as written _compiles_ with earlier versions of `tasty` as well.

Otherwise looks good, please add a changelog entry (check `tasty-quickcheck` changelog for suggested format).

michaelpj:
Oh, of course, the callback was there before it just didn't do anything.

michaelpj:
Okay, updated.

Bodigrim (approved):

andreasabel (approved):
So the progress reporting is a feature controlled by `tasty`; `tasty-unit` only enables it here.
Seems fine with the PVP.  (Caused me a bit of headache thinking about this case.)

PR #389, opened 2023-09-16
Make computeAlignment more efficient

Bodigrim:
Currently `Test.Tasty.Ingredients.ConsoleReporter.computeAlignment` evaluation is extremely inefficient: firstly because `data Maximum` lacks strictness annotations, and secondly because `foldTestTree` operates not even over `Maximum a`, but over `Int -> Maximum a`.

The patch improves efficiency of `computeAlignment` and arguably makes its implementation simpler.

CC @martijnbastiaan for review.

### DIFF
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -749,35 +749,19 @@ formatDesc n desc =
       then paddedDesc
       else chomped
 
-data Maximum a
-  = Maximum a
-  | MinusInfinity
-
-instance Ord a => Sem.Semigroup (Maximum a) where
-  Maximum a <> Maximum b = Maximum (a `max` b)
-  MinusInfinity <> a = a
-  a <> MinusInfinity = a
-instance Ord a => Monoid (Maximum a) where
-  mempty = MinusInfinity
-#if !MIN_VERSION_base(4,11,0)
-  mappend = (Sem.<>)
-#endif
-
 -- | Compute the amount of space needed to align \"OK\"s and \"FAIL\"s
 computeAlignment :: OptionSet -> TestTree -> Int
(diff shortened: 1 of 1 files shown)


core/Test/Tasty/Ingredients/ConsoleReporter.hs
```diff
@@ -749,35 +749,19 @@ formatDesc n desc =
...
-      { foldSingle = \_ name _ level -> Maximum (stringWidth name + level)
-      , foldGroup = \_opts _ m -> mconcat m . (+ indentSize)
+  max 0 .
+  foldTestTree0
+    minBound
+    TreeFold
+      { foldSingle = \_ name _ -> stringWidth name
+      , foldGroup = \_ _ m -> if null m then minBound else maximum m + indentSize
+      , foldResource = \_ _ f -> f $ throwIO NotRunningTests
+      , foldAfter = \_ _ _ b -> b
```
VictorCMiraldo, line 762: I think we could make this patch even smaller borrowing the `trivialFold` from the original function:

```suggestion
    trivialFold
      { foldSingle = \_ name _ -> stringWidth name
      , foldGroup = \_ _ m -> if null m then minBound else maximum m + indentSize
```

VictorCMiraldo (approved):
LGTM! Thanks for the patch @Bodigrim

core/Test/Tasty/Ingredients/ConsoleReporter.hs
Bodigrim: `trivialFold` unfortunately requires `Monoid` constraint, which we don't have here.

VictorCMiraldo: Too bad; ignore my suggestion then :)

martijnbastiaan (approved):
Looks much simpler!

PR #390, opened 2023-09-24
Two IntMap-related improvements

Bodigrim:
* `launchTestTree`: use `IntMap` strict in values

  At the moment we use `Data.IntMap` which is lazy in values, and we initialise values with thunks `testStatus <$> testActions`. This is outright silly: `testStatus` is just a selector for `TVar` field, forcing it to WHNF can do no harm.

  We also take an opportunity to replace `IntMap.fromList` with `IntMap.fromDistinctAscList`.

* `statusMapResult`: use `IntSet` instead of `IntMap ()`

  It seems the only reason behind `IntMap ()` was the absense of `Data.IntMap.withoutKeys`, which was only added in `containers-0.5.8`.

CC @martijnbastiaan for review

### DIFF
--- a/core/tasty.cabal
+++ b/core/tasty.cabal
@@ -61,7 +61,7 @@ library
   build-depends:
     base                 >= 4.9  && < 5,
     stm                  >= 2.3  && < 2.6,
-    containers                      < 0.7,
+    containers           >= 0.5.8 && < 0.7,
     transformers         >= 0.5  && < 0.7,
     tagged               >= 0.5  && < 0.9,
     optparse-applicative >= 0.14 && < 0.19,
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -50,6 +50,7 @@ import Test.Tasty.Runners.Reducers
 import Test.Tasty.Runners.Utils
 import Text.Printf
 import qualified Data.IntMap as IntMap
+import qualified Data.IntSet as IntSet
 import Data.Char
 #ifdef USE_WCWIDTH
 import Foreign.C.Types (CInt(..), CWchar(..))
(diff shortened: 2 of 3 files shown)


core/tasty.cabal
```diff
@@ -61,7 +61,7 @@ library
   build-depends:
     base                 >= 4.9  && < 5,
     stm                  >= 2.3  && < 2.6,
-    containers                      < 0.7,
```
andreasabel, line 64: GHC 8.0 ships `containers-0.5.7` according to https://downloads.haskell.org/~ghc/8.0.2/docs/html/users_guide/8.0.1-notes.html#containers, so maybe this bump is a little too early?

The CI here does not catch this. Haskell CI would because it works with `constraints: containers installed`.

(GHC 8.2 then ships with 0.5.10.)

andreasabel (commented):
I suggest to hold the changes that need a bump to `containers` until we drop GHC 8.0.

core/tasty.cabal
```diff
@@ -61,7 +61,7 @@ library
...
+    containers           >= 0.5.8 && < 0.7,
```
andreasabel, line 64: GHC 8.0 ships `containers-0.5.7` according to https://downloads.haskell.org/~ghc/8.0.2/docs/html/users_guide/8.0.1-notes.html#containers, so maybe this bump is a little too early?

The CI here does not catch this. Haskell CI would because it works with `constraints: containers installed`.

(GHC 8.2 then ships with 0.5.10.)

Bodigrim:
@andreasabel good catch. I can put CPP of course, but I have no reason to keep compatibility with GHC < 8.6 (where 8.6 is only because of GHCJS). How do you feel about dropping earlier versions?

andreasabel:
Thinking it over, I guess it is fine to go ahead with this PR as-is.
Folks that insist on `constraints: containers installed` in their CI can drop GHC 8.0 from the testing phase there.
For everyone else, this is not a problem.

martijnbastiaan (approved):
_Generally_ not a fan of `fromDistinctAscList` as it is easy to break the precondition, but given that the ascending list is right behind it this LGTM.

andreasabel (approved):

VictorCMiraldo:
FWIW, looks good to me too! :) Thank you for the improvements!

PR #398, opened 2023-10-28
Avoid blanket import from `GHC.Conc`

edsko:
Corresponding ticket for `async` itself at https://github.com/simonmar/async/pull/153.

### DIFF
--- a/core/Control/Concurrent/Async.hs
+++ b/core/Control/Concurrent/Async.hs
@@ -48,7 +48,7 @@ import Control.Concurrent
 import Control.Monad
 import Data.IORef
 import Data.Typeable
-import GHC.Conc
+import GHC.Conc (ThreadId(..))
 import GHC.Exts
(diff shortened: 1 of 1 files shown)


Bodigrim:
@amesgen if you get a moment to fix WASM CI job, it would be most appreciated. (I assume it's just a hash change, but cannot verify AFK)

Bodigrim:
@edsko could you please rebase?

Any particular motivation for this change? Why `GHC.Conc` only?

edsko:
@Bodigrim Have rebased. As for motivation: `GHC.Conc` exports some stuff that `Control.Concurrent.STM` also exports, which leads to name clashes when using a version of `stm` that doesn't straight up re-export stuff from `GHC.Conc` but does something slightly different (such as specifically https://github.com/edsko/stm-debug).

Bodigrim:
Thanks!

PR #403, opened 2023-12-06
Improve --hide-successes to not print headings for empty test subtrees

sergv:
Improve --hide-successes to not print headings for empty test subtrees (i.e. when -p filtered tests out).

This is again inspired by Emacs which cannot deal with escape sequence that moves cursor up one line. This change makes those movement unneeded since heading is not printed if there are no tests under it.

Should benefit everyone though since there's no point in printing only to delete afterwards, even if one wants to see progress - there's no progress being made in those empty subtrees that were filtered out.


sergv:
Updated with different approach: instead of filtering out empty subtrees during printing, don't generate them in the first place in the `foldTestTree`.

Bodigrim:
@martijnbastiaan could you possibly review this? I think it touches parts of `tasty` which you were involved with recently.

martijnbastiaan (approved):
LGTM!

PR #404, opened 2023-12-14
Mention the proper env vars for usage in MinGW

jasagredo:
If one doesn't set that flag, by default any argument `/.../` gets converted to a path by MSYS2 ([see](https://www.msys2.org/docs/filesystem-paths/)). 

```
❯ cabal run my-testsuite -- -p "/my pattern/"
option -p: Could not parse: C:/msys64/my pattern/ is not a valid pattern
...
❯ MSYS2_ARG_CONV_EXCL=* cabal run my-testsuite -- -p "/my pattern/"
<it works>
```

This will print the following advise on failing tests when running on what looks like a MinGW enviroment:

```
          Use -p '/my pattern/' to rerun this test only.
          If running in a MinGW shell, set the environment variable 'MSYS2_ARG_CONV_EXCL=*' or the pattern above will be mangled.
```

It still uses the "If" formulation because PowerShell running the mingw haskell toolchain won't need this parameter (modulo haskell/cabal#9519):

```
PS C:\Users\Javier\example> cabal run my-testsuite -- -p "/my pattern/"
<works>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -820,9 +820,12 @@ Use `sequentialTestGroup` to mitigate these problems.
 
 3. **Q**: Patterns with slashes do not work on Windows. How can I fix it?
 
-   **A**: If you are running Git for Windows terminal, it has a habit of converting slashes
-   to backslashes. Set `MSYS_NO_PATHCONV=1` to prevent this behaviour, or follow other
-   suggestions from [Known Issues](https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.md#known-issues).
+   **A**: If you are running Git for Windows terminal, it has a habit of
+   converting slashes to backslashes. Set `MSYS_NO_PATHCONV=1` when running the
+   Git for Windows terminal and `MSYS2_ARG_CONV_EXCL=*` when running a MinGW
+   bash directly to prevent this behaviour, or follow other suggestions from
+   [Known
+   Issues](https://github.com/git-for-windows/build-extra/blob/main/ReleaseNotes.md#known-issues).
 
 ## Press
 
--- a/core/Test/Tasty/Patterns.hs
+++ b/core/Test/Tasty/Patterns.hs
@@ -48,7 +48,15 @@ instance IsOption TestPattern where
   defaultValue = noPattern
   parseValue = parseTestPattern
   optionName = return "pattern"
-  optionHelp = return "Select only tests which satisfy a pattern or awk expression"
+#if !defined(mingw32_HOST_OS)
+  optionHelp = return "Select only tests which satisfy a pattern or awk expression."
+#else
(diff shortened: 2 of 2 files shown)


Bodigrim:
[FAQ](https://github.com/UnkindPartition/tasty#faq) recommends setting `MSYS_NO_PATHCONV=1`. Is `MSYS2_ARG_CONV_EXCL=*` better or different?

It would be better to add the instruction to the error message for failed parse instead of repeating it after each failed test.

jasagredo:
I was unable to make it work with `MSYS_NO_PATHCONV` but I can try again.

And I agree it would be better to do it in the parse error message. Good point, will do.

jasagredo:
```
❯ MSYS_NO_PATHCONV=1 cabal run cardano-test -- -p "/ExtLedgerState_Conway/"
option -p: Could not parse: C:/msys64/ExtLedgerState_Conway/ is not a valid pattern
...

❯ MSYS2_NO_PATHCONV=1 cabal run cardano-test -- -p "/ExtLedgerState_Conway/"
option -p: Could not parse: C:/msys64/ExtLedgerState_Conway/ is not a valid pattern
...
```

jasagredo:
As the error in parsing is created by `mkOptionCLParser` using just the name of the parser, in order to not print this error for all parsers (which wouldn't make sense), would it be better to add a field to `IsOption v` something like:
```haskell
class Typeable v => IsOption v where
  -- | Optional additional hints to present on the CLI when this option 
  -- fails to parse, for example OS specific hints.
  additionalHints :: Tagged v (Maybe String)
```
?

Bodigrim:
Maybe we can unmangle automatically what MSYS has mangled? That would be the best solution.

jasagredo:
I don't think that would be a good approach, I think tasty should try to interpret the input as a pattern always, and fail if it is not a pattern. 

At most the slashes could be optional maybe? That would avoid MSYS2 mangling the parameter.

(The above is my current opinion, I don't know the internals of tasty or the overall design and goals so I could be wrong)

Bodigrim:
`MSYS_NO_PATHCONV` used to work for me in MinGW shells like Git Bash. It’s surprising that it does not work for you, has something changed on MSYS side?..

There are two low-hanging fruits: we can update `README.md` with more robust instructions and we can also extend `—help` output wrt `—pattern`. Other than that I’d prefer to extend the parser of patterns to demangle things: after all a user might not be in position to manipulate environment variables at all.

jasagredo:
Let's just update the README and let it be.

jasagredo:
Let me know if this looks good now.

Bodigrim (approved):

Bodigrim:
Thanks!

PR #410, opened 2024-03-18
Avoid running earlier successes with --quickcheck-replay

### LINKED ISSUE #383: QuickCheck: allow to print/specify seed for a failure without having to replay through previous successes
If we find a property test failure after a lot of iterations, reproducing it is annoying as we have to replay all successful iterations to get to the failing test case. This is a bit annoying:

 - major: reproducing the failure can take a long time when testing with lots of iterations and/or long test durations per test (especially true for long-running nightly tests) 
 - minor: ad-hoc debugging statements via eg `Debug.Trace` are also printed for all previous successful cases

Also see https://github.com/hedgehogqa/haskell-hedgehog/pull/454 for a recent Hedgehog PR implementing this (and more).

I think this should be simple to implement in tasty-quickcheck as QuickCheck exposes the necessary seed in its [`Result`](https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#t:Result):

```haskell
data Result
  = ...
  | Failure
    { ...
    , usedSeed        :: QCGen
      -- ^ What seed was used
    , usedSize        :: Int
      -- ^ What was the test size
    ...
    }
...

data Args
  = Args
  { replay          :: Maybe (QCGen,Int)
    -- ^ Should we replay a previous test?
  , ...
  }
```

On failure, `(usedSeed, usedSize)` would be printed, and it could be specified via a new flag `--quickcheck-exact-replay` (name very tentative).

Happy to open a PR if this sounds like a useful addition.

facundominguez:
On failure, tasty-quickcheck now suggests string seeds (with an implementation-dependent structure). Integer seeds are still accepted as input though.

Fixes #383

### DIFF
--- a/quickcheck/CHANGELOG.md
+++ b/quickcheck/CHANGELOG.md
@@ -4,6 +4,9 @@ Changes
 Next Version
 ------------
 
+Version 0.11
+--------------
+
 * Produce seeds that run a single failing tests instead of reproducing
   all the earlier successes ([#410](https://github.com/UnkindPartition/tasty/pull/410)).
 
--- a/quickcheck/CHANGELOG.md
+++ b/quickcheck/CHANGELOG.md
@@ -7,7 +7,7 @@ Next Version
 * Produce seeds that run a single failing tests instead of reproducing
   all the earlier successes ([#410](https://github.com/UnkindPartition/tasty/pull/410)).
 
-  Seeds are now strings, instead of single integers. e.g.
+  Seeds are now strings, instead of single integers, e.g.
   `--quickcheck-replay="(SMGen 2909028190965759779 12330386376379709109,0)"`
 
   Single integer seeds are still accepted as input, but they do run through
--- a/quickcheck/CHANGELOG.md
+++ b/quickcheck/CHANGELOG.md
@@ -1,6 +1,18 @@
 Changes
 =======
 
+Next Version
+------------
+
+* Produce seeds that run a single failing tests instead of reproducing
+  all the earlier successes ([#410](https://github.com/UnkindPartition/tasty/pull/410)).
+
+  Seeds are now strings, instead of single integers. e.g.
+  `--quickcheck-replay="(SMGen 2909028190965759779 12330386376379709109,0)"`
+
+  Single integer seeds are still accepted as input, but they do run through
+  earlier successes.
+
 Version 0.10.3
 --------------
 
--- a/quickcheck/CHANGELOG.md
+++ b/quickcheck/CHANGELOG.md
@@ -7,11 +7,16 @@ Next Version
 * Produce seeds that run a single failing tests instead of reproducing
   all the earlier successes ([#410](https://github.com/UnkindPartition/tasty/pull/410)).
 
-  Seeds are now strings, instead of single integers, e.g.
+  Seeds are now pairs instead of single integers, e.g.
   `--quickcheck-replay="(SMGen 2909028190965759779 12330386376379709109,0)"`
 
   Single integer seeds are still accepted as input, but they do run through
   earlier successes.
+
+  The `QuickCheckReplay` type used as a tasty option has three data constructors
+  now. `QuickCheckReplayNone` is the default value and provides no seed.
+  `QuickCheckReplayLegacy` takes an integer as before. The `QuickCheckReplay`
+  data constructor takes the new seed form.
 
 Version 0.10.3
 --------------
--- a/quickcheck/Test/Tasty/QuickCheck.hs
+++ b/quickcheck/Test/Tasty/QuickCheck.hs
@@ -85,11 +85,17 @@ newtype QuickCheckTests = QuickCheckTests Int
 -- | Replay seed
 data QuickCheckReplay
     = -- | No seed
+      --
+      -- @since 0.11
       QuickCheckReplayNone
     | -- | Legacy integer seed
+      --
+      -- @since 0.11
       QuickCheckReplayLegacy Int
     | -- | @(qcgen, intSize)@ holds both the seed and the size
       -- to run QuickCheck tests
+      --
+      -- @since 0.11
       QuickCheckReplay (QCGen, Int)
   deriving (Typeable)
 
--- a/quickcheck/Test/Tasty/QuickCheck.hs
+++ b/quickcheck/Test/Tasty/QuickCheck.hs
@@ -50,16 +50,16 @@ import Test.QuickCheck hiding -- for re-export
   , verboseCheckAll
   )
 
+import Control.Applicative
 import qualified Data.Char as Char
 import Data.Typeable
 import Data.List
 import Text.Printf
 import Text.Read (readMaybe)
-import Test.QuickCheck.Random (mkQCGen)
+import Test.QuickCheck.Random (QCGen, mkQCGen)
 import Options.Applicative (metavar)
 import System.Random (getStdRandom, randomR)
 #if !MIN_VERSION_base(4,9,0)
-import Control.Applicative
 import Data.Monoid
 #endif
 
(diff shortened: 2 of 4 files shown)


quickcheck/Test/Tasty/QuickCheck.hs
```diff
@@ -169,34 +179,41 @@ instance IsOption QuickCheckMaxShrinks where
 -- but may be used by others.
 --
 -- @since 0.9.1
-optionSetToArgs :: OptionSet -> IO (Int, QC.Args)
+optionSetToArgs :: OptionSet -> IO QC.Args
```
facundominguez, line 182: Notice the API breaking change. `optionSetToArgs` is exported.

quickcheck/Test/Tasty/QuickCheck.hs
```diff
@@ -259,9 +277,12 @@ successful r =
...
 
-makeReplayMsg :: Int -> Int -> String
-makeReplayMsg seed size = let
-    sizeStr = if (size /= defaultMaxSize)
-                 then printf " --quickcheck-max-size=%d" size
-                 else ""
-  in printf "Use --quickcheck-replay=%d%s to reproduce." seed sizeStr
+makeReplayMsg :: (QCGen, Int) -> String
+makeReplayMsg seedSz =
+  printf "Use --quickcheck-single-replay=\"%s\" to reproduce." (show seedSz)
```
facundominguez, line 282: Probably we should display both the value of `--quickcheck-replay` and `--quickcheck-single-replay` here. Otherwise, the user of `--quickcheck-replay` has no way to get a seed.

quickcheck/Test/Tasty/QuickCheck.hs
Bodigrim: I don't see a reason for breakage though. You can still return `(Int, QC.Args)`.

quickcheck/Test/Tasty/QuickCheck.hs
Bodigrim: What are the remaining use cases for `--quickcheck-replay` now? I guess you can just reuse it without introducing a new command line flag.

quickcheck/Test/Tasty/QuickCheck.hs
facundominguez: Indeed. I reverted this.

quickcheck/Test/Tasty/QuickCheck.hs
facundominguez: I don't know who is relying on the old behavior. I'm happy to break that if nobody cares.

Bodigrim: I find it hard to imagine that someone would consciously prefer old behaviour to a new, but it seems you can support both without too much trouble. If it's `--quickcheck-replay=(QCGen ..., 42)` do a new thing, if it's `--quichckeck-replay=QCGen...` do the old thing.

facundominguez: And when printing, should we show both forms when sensible?

Bodigrim: Showing one form would be enough IMO.

facundominguez: It can be done as you suggest. But note that the old form is unusable if the user has no way to discover the seeds that go with it. What would be the purpose of keeping the old form then?

Bodigrim: Just an escape hatch. I do not see why a user might want the old behavior back; but someone could have a regression script still using an old format. Not crashing in such scenario would be useful.

facundominguez: Pushed. Also added a note to the change log. Thanks for your review @Bodigrim!

quickcheck/CHANGELOG.md
```diff
@@ -1,6 +1,18 @@
 Changes
 =======
 
+Next Version
+------------
+
+* Produce seeds that run a single failing tests instead of reproducing
+  all the earlier successes ([#410](https://github.com/UnkindPartition/tasty/pull/410)).
+
+  Seeds are now strings, instead of single integers. e.g.
```
Bodigrim, line 10:
```suggestion
  Seeds are now strings, instead of single integers, e.g.
```

quickcheck/Test/Tasty/QuickCheck.hs
```diff
@@ -82,7 +82,10 @@ testProperties name = testGroup name . map (uncurry testProperty)
 newtype QuickCheckTests = QuickCheckTests Int
   deriving (Num, Ord, Eq, Real, Enum, Integral, Typeable)
 
-newtype QuickCheckReplay = QuickCheckReplay (Maybe Int)
+-- | Replay seed. The @Left int@ form is kept for legacy purposes.
+-- The form @Right (qcgen, intSize)@ holds both the seed and the size
+-- to run QuickCheck tests.
+newtype QuickCheckReplay = QuickCheckReplay (Maybe (Either Int (QCGen, Int)))
```
Bodigrim, line 88: Could it be
```suggestion
newtype QuickCheckReplay = QuickCheckReplay (Maybe (Int, Maybe QCGen))
```
?

quickcheck/Test/Tasty/QuickCheck.hs
```diff
@@ -168,30 +173,37 @@ instance IsOption QuickCheckMaxShrinks where
 -- This is a low-level function that was originally added for tasty-hspec
 -- but may be used by others.
 --
+-- The returned Int is kept only for backward compatibility purposes. It
+-- has no use in tast-quickcheck.
```
Bodigrim, line 177:
```suggestion
-- has no use in @tasty-quickcheck@.
```

Bodigrim (commented):
A few more wibbles, but overall looks good to me, thanks!

CC @andreasabel @martijnbastiaan @VictorCMiraldo to review.

quickcheck/Test/Tasty/QuickCheck.hs
facundominguez: I find the code easier to approach as it is.

* The two Ints in the Either type have different meanings, whether the meaning of the Int in the Maybe form depends on the value of the second component of the pair.

* The quickcheck API can take the pair `(QCGen, Int)` as is without further ado.

But let me know if you still think the `Maybe` form is to be preferred.

Another option is to use a custom type:
```haskell
data Seed = OldSeed Int | NewSeed (QCGen, Int)
```

Bodigrim: OK, unless someone else feels strongly about it, I'm fine to keep it as is.

facundominguez:
Thanks @martijnbastiaan 

> I do like a custom data type better (Seed/QuickCheckReplay),

I contributed an implementation of this idea in https://github.com/facundominguez/tasty/pull/1.
I could make it a follow up PR, or maintainers can cherry-pick it if they wish.

Bodigrim:
> I contributed an implementation of this idea in [facundominguez#1](https://github.com/facundominguez/tasty/pull/1).
> I could make it a follow up PR to `UnkindPartition/tasty` or maintainers can cherry-pick it if they wish.

This looks nice. Please add it to this PR + mention the change to `QuickCheckReplay` in the changlog + bump major version of `tasty-quickcheck` - and I think we'll be good to merge.

Bodigrim:
+ please ensure that everything has correct `@since` pragmas.

facundominguez:
> - [x] add it to this PR
> - [x] mention the change to `QuickCheckReplay` in the changelog
> - [x] bump major version of `tasty-quickcheck`

> - [ ] please ensure that everything has correct `@since` pragmas.

I added "@since 0.11" tags to the data constructors of the `QuickCheckReplay` data type. Let me know if more tags are needed.

Bodigrim:
@facundominguez thanks! `README.md` can use an update too, if you are up to it.

@amesgen as an original author of #383, are you happy with this solution?

facundominguez:
> README.md can use an update too, if you are up to it.

I'm a bit embarrassed to admit that I cannot discover on my own what needs updating there. :facepalm:

amesgen (approved):
This looks excellent!

Bodigrim:
> I'm a bit embarrassed to admit that I cannot discover on my own what needs updating there. 🙈

@facundominguez the output of `--help` could be updated, but I'm not sure it was kept up-to-date before anyway. 

Thanks for your contribution, merging.

PR #412, opened 2024-04-06
Add ability to supply options programmatically

newhoggy:


Bodigrim:
@newhoggy sorry, but I would really love if we first have an issue to discuss the problem and alternative solutions before jumping straight to a PR. Could it be solved without extending an ever-growing API?

newhoggy:
https://github.com/UnkindPartition/tasty/issues/414

Issue #414, opened 2024-04-08
Add ability to supply options programmatically


newhoggy:
Sometimes it is useful for a test suite to use a default that is different than the one that `tasty` uses.

For example a test suite might have some concurrency issues and needs `NumThreads 1` to be able to run correctly.

Bodigrim:
Are existing mechanisms (such as [`adjustOption`](https://hackage.haskell.org/package/tasty-1.5/docs/Test-Tasty.html#v:adjustOption)) insufficient? Why?

There is also [`sequentialTestGroup`](https://hackage.haskell.org/package/tasty-1.5/docs/Test-Tasty.html#v:sequentialTestGroup) which can potentially mitigate multithreading issues as discussed in #406.

Bodigrim:
If this is about `NumThreads` specifically, `defaultMain*` could unwrap top levels of `PlusTestOptions` and incorporate them into `opts`. This would avoid extending API.

newhoggy:
`sequentialTestGroup` doesn't work for us because it creates dependencies between tests.

This breaks `--pattern` for us because if we select a test with `--pattern` it will run not just the selected tests but its dependencies as well.

Bodigrim:
https://github.com/UnkindPartition/tasty/blob/83b3a53087b391bf831c19e0a04dd2dd93dbb884/core/Test/Tasty/Run.hs#L605-L607

Instead of directly going for `lookupOption`, one can unwrap top-level `PlusTestOptions` from `tree` and apply them to `opts` first, then lookup `NumThreads` in the result.

newhoggy:
Hmm, should it not be other way around?  ie. if `--num-threads` is supplied on the command line, it should overwrite what's specified by the tests?

newhoggy:
Is `tasty` expressive enough to model `SingleThreaded` as an option?

That would be ideal for us because it would allow a sub-tree to be run single threaded allowing the rest to be parallel.

newhoggy:
There are some other things I'm also interest in:

* A way to time tests and record them in a file.
* A way to put timeouts on a test
* A way to retry a test if it fails
* A way to retry a test if it times out

Bodigrim:
> Is `tasty` expressive enough to model `SingleThreaded` as an option?
> 
> That would be ideal for us because it would allow a sub-tree to be run single threaded allowing the rest to be concurrent.

[...]

Issue #406, opened 2024-02-03
It would be nice to be able to limit the number of jobs used by `TestTree`s


michaelpj:
From looking at the code and experimentation, I _believe_ that the `NumThreads` option only takes effect at the top-level. So you can't, for example, limit a particular test tree to be run with only one job (in case the tests interfere with each other).

michaelpj:
And that led me to `ExecutionMode`, which seems like it's exactly what I want but sadly isn't exported. Maybe there's a reason why it's only available through `sequentialTestGroup`, but I'm not sure :thinking:

michaelpj:
Indeed, `sequentialTestGroup` does not do what I want, since it only makes the tests directly in the group run sequentially, whereas tests in nested groups continue to execute in parallel. I really do just want to set the execution mode to sequential for a whole test tree!

Bodigrim:
@michaelpj it is admittedly awkward, but you can pattern-match on 
```haskell
PlusTestOptions mkSequential = sequentialTestGroup "foo" AllFinish []
```
and then
```haskell
defaultMain $ PlusTestOptions mkSequential testTree
```

I'm curious if it works, please tell if you give it a try.

dyniec:
contructors of TestTree aren't exported, so PlusTestOptions is not visible :/

Bodigrim:
@dyniec constructors of `TestTree` are exported from https://hackage.haskell.org/package/tasty-1.5/docs/Test-Tasty-Runners.html#t:TestTree

PR #417, opened 2024-04-29
Add ability to supply options for launchers and reporters at the top-level of test tree

Bodigrim:
While tasty allows to configure `OptionSet` at any level of test tree, it often has any effect only on options of test providers (`class IsTest`). But test runners and reporters typically only look into the `OptionSet` they were given as an argument. This is not unreasonable: e. g., if an option is a log filename you cannot expect to change it in the middle of the run. It is however too restrictive: there is no way to use `defaultMain` but hardcode a global option, without passing it via command line.

`applyTopLevelPlusTestOptions` allows for a compromise: unwrap top-level `PlusTestOptions` from the `TestTree` and apply them to the `OptionSet` from command line. This way a user can wrap their tests in `adjustOption` / `localOption` forcing, for instance, `NumThreads` to 1.

Related to #406 and #414.

### DIFF
--- a/cabal.project
+++ b/cabal.project
@@ -14,3 +14,5 @@ if os(wasi)
   package splitmix
     tests: False
     benchmarks: False
+  constraints:
+    hashable -arch-native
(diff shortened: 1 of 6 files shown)


Bodigrim:
This is ready for reivew, CC @martijnbastiaan.

martijnbastiaan (approved):
Seems sensible! Thanks for tagging, this one slipped by.

martijnbastiaan:
Oh, changelog + `since` annotation perhaps?

Bodigrim:
> Oh, changelog + `since` annotation perhaps?

Done, thanks.

PR #419, opened 2024-05-05
fixed quickcheck percent done error

### LINKED ISSUE #402: Progress reporting for QuickCheck seems to report progress over 100%
I'm not really sure what's going on here, but I've seen progress reports from QuickCheck go over 100%. These are for tests where I've set `MaxSuccesses` to a higher number, so initially I thought we simply weren't dividing by the right thing. But the code [seems to be doing the right thing](https://github.com/UnkindPartition/tasty/blob/master/quickcheck/Test/Tasty/QuickCheck.hs#L251) so I'm not sure what's happening.

cc @coot

BebeSparkelSparkel:
closes #402

### DIFF
--- a/quickcheck/Test/Tasty/QuickCheck.hs
+++ b/quickcheck/Test/Tasty/QuickCheck.hs
@@ -258,7 +258,7 @@ quickCheck :: (Progress -> IO ())
            -> IO QC.Result
 quickCheck yieldProgress args
   = (.) (QC.quickCheckWithResult args)
-  $ QCP.callback 
+  $ QCP.callback
   $ QCP.PostTest QCP.NotCounterexample
   $ \QC.MkState {QC.maxSuccessTests, QC.numSuccessTests} _ ->
     yieldProgress $ emptyProgress {progressPercent = fromIntegral numSuccessTests / fromIntegral maxSuccessTests}
--- a/quickcheck/Test/Tasty/QuickCheck.hs
+++ b/quickcheck/Test/Tasty/QuickCheck.hs
@@ -24,9 +24,8 @@ import Test.Tasty ( testGroup )
 import Test.Tasty.Providers
 import Test.Tasty.Options
 import qualified Test.QuickCheck as QC
-import qualified Test.QuickCheck.Test as QC
+import qualified Test.QuickCheck.Property as QCP
 import qualified Test.QuickCheck.State as QC
-import qualified Test.QuickCheck.Text as QC
 import Test.Tasty.Runners (formatMessage, emptyProgress)
 import Test.QuickCheck hiding -- for re-export
   ( quickCheck
@@ -51,21 +50,15 @@ import Test.QuickCheck hiding -- for re-export
   )
 
 import Control.Applicative
-import qualified Data.Char as Char
 import Data.Typeable
-import Data.IORef (newIORef, readIORef, writeIORef)
 import Data.List
 import Text.Printf
-import Text.Read (readMaybe)
 import Test.QuickCheck.Random (QCGen, mkQCGen)
 import Options.Applicative (metavar)
 import System.Random (getStdRandom, randomR)
 #if !MIN_VERSION_base(4,9,0)
 import Data.Monoid
 #endif
-import GHC.IO.Unsafe (unsafePerformIO)
-import Data.Functor (($>))
-import Test.QuickCheck.Property (mapResult, maybeNumTests)
 
 newtype QC = QC QC.Property
   deriving Typeable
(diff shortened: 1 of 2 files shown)


BebeSparkelSparkel:
@Bodigrim Any thoughts on pulling this fix?

Bodigrim (commented):
I don't know internals of `QuickCheck` well enough. @MaximilianAlgehed could you possibly review?

quickcheck/Test/Tasty/QuickCheck.hs
```diff
@@ -259,14 +263,21 @@ quickCheck :: (Progress -> IO ())
...
+  maxSuccess <- newIORef $ fromIntegral argMaxSuccess
+  let setMaxTests r = case maybeNumTests r of
+        Just n -> unsafePerformIO $ writeIORef maxSuccess (fromIntegral n) $> r
+        Nothing -> r
   tm <- QC.newTerminal
           (const $ pure ())
-          (\progressText -> yieldProgress emptyProgress { progressPercent = parseProgress progressText })
+          $ \progressText -> do
+            ms <- readIORef maxSuccess
+            yieldProgress emptyProgress { progressPercent = 100 * parseProgress progressText / ms }
```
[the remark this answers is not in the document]
MaximilianAlgehed: Assuming `parseProgress` returns the number of completed tests this code is "fine" as the 100 is just there to convert from fraction to percent.

The bigger issue here is that you can also set the number of tests with `withMaxSuccess` and this code doesn't capture that. However, it's a bit tricky to deal with that given the current interface to QuickCheck.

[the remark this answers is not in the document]
BebeSparkelSparkel: `parseProgress` does just returns the number of tests run.

"this code doesn't capture that" do you mean the current implementation or the changes that I have made?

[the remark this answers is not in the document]
MaximilianAlgehed: I don't think neither the changes you've made or the original code captures the fact that `withMaxSuccess` changes the `Result` type to change the number of tests. The best way I see for you to account for that change is to use a `callback` in QuickCheck and update an `IORef` when you see the `maybeMaxSuccess` field change in the `Result` argument to your callback.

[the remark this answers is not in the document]
BebeSparkelSparkel: I have tried this and the changes I have made do indeed fix the problem, but this may not be the optimum solution.

I am unfamiliar with `callback` could you give me a link describing how to use it?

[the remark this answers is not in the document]
MaximilianAlgehed: Ok, if this works then there is no point doing it differently.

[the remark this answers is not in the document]
BebeSparkelSparkel: Does the callback remove the need for unsafePerformIO?

MaximilianAlgehed:
Probably yes.

Den ons 29 maj 2024 16:41William Rusnack ***@***.***> skrev:

BebeSparkelSparkel:
Then it would probably be worth changing

MaximilianAlgehed:
Callbacks get both the state and the result so it should be.

Bodigrim:
@BebeSparkelSparkel do you need any help to make further progress here? I'd like to make a release of `tasty` and stuff in upcoming weeks.

BebeSparkelSparkel:
Yes, I do not know how to test this besides the visual I added.

Also, with callbacks should shrinks be reported in the progress?

quickcheck/tests/test.hs
```diff
@@ -107,6 +109,9 @@ main =
           resultDescription =~ "Failed.*expected failure"
           resultDescription =~ "Use .* to reproduce"
 
+      -- ???????? not sure how to test intermediate results
```
Bodigrim, line 112:
```suggestion
      -- Run the test suite manually and check that progress does not go beyond 100%
```

quickcheck/Test/Tasty/QuickCheck.hs
```diff
@@ -263,28 +256,12 @@ quickCheck :: (Progress -> IO ())
...
-    parseProgress = maybe 0 (\n -> fromIntegral (n :: Int) / fromIntegral (QC.maxSuccess args))
-                  . readMaybe
-                  . takeWhile Char.isDigit
-                  . drop 1
+quickCheck yieldProgress args
+  = (.) (QC.quickCheckWithResult args)
+  $ QCP.callback 
+  $ QCP.PostTest QCP.NotCounterexample
+  $ \QC.MkState {QC.maxSuccessTests, QC.numSuccessTests} _ ->
+    yieldProgress $ emptyProgress {progressPercent = fromIntegral numSuccessTests / fromIntegral maxSuccessTests}
```
Bodigrim, line 264: Wow. This looks awesome, much better than before.

quickcheck/Test/Tasty/QuickCheck.hs
```diff
@@ -259,21 +256,12 @@ quickCheck :: (Progress -> IO ())
...
-  where
-    -- QuickCheck outputs something like "(15461 tests)\b\b\b\b\b\b\b\b\b\b\b\b\b"
-    parseProgress :: String -> Float
-    parseProgress = maybe 0 (\n -> fromIntegral (n :: Int) / fromIntegral (QC.maxSuccess args))
-                  . readMaybe
-                  . takeWhile Char.isDigit
-                  . drop 1
+quickCheck yieldProgress args
+  = (.) (QC.quickCheckWithResult args)
+  $ QCP.callback
```
Bodigrim, line 261:
```suggestion
  $ QCP.callback
```

Bodigrim:
@BebeSparkelSparkel looks good to me modulo two suggestions above.

quickcheck/tests/test.hs
BebeSparkelSparkel: @MaximilianAlgehed Do you have any ideas on how to test this properly?

Bodigrim: We don't really have UI tests of this sort, so let's not overthink it much.

quickcheck/Test/Tasty/QuickCheck.hs
BebeSparkelSparkel: One thing that bothers me is that shrinks and discards are not reported. `callback` makes this data very accessible with [State](https://github.com/nick8325/quickcheck/blob/7e972cadfe9745db47a0e625c20096ec854da7f8/src/Test/QuickCheck/State.hs#L18) and I think these should be reported.

Bodigrim: It's important to keep progress text short, otherwise TUI is likely to break. I don't think there is space to fit information about discarded / shrinks. 

Let's not scope creep please, I have very limited resource to review things here.

BebeSparkelSparkel:
Done pull

Bodigrim (approved):

Bodigrim:
Thanks, that's great.

PR #420, opened 2024-06-15
Release tasty-1.5.1, tasty-quickcheck-0.11, tasty-hunit-0.10.2

Bodigrim:


VictorCMiraldo (approved):
:+1: Thank you!

Bodigrim:
Released:
* https://hackage.haskell.org/package/tasty-1.5.1
* https://hackage.haskell.org/package/tasty-quickcheck-0.11
* https://hackage.haskell.org/package/tasty-hunit-0.10.2

Bodigrim:
Apparently changes to cursor manipulations broke things:
```
All
  read:          OK8  ms ±  23 ms
    278  ms ±  23 ms
  write:         OK0  ms ±  38 ms
    900  ms ±  38 ms
  index build:   OK1  ms ±  15 ms
    521  ms ±  15 ms
  index rebuild: OK0  ms ± 7.5 ms
    110  ms ± 7.5 ms
```

I revised `tasty-1.5.1` to decomission it for time being: https://hackage.haskell.org/package/tasty-1.5.1/revisions/

Bodigrim:
Apparently reverting b152a0bc63166a4592e1f3639ef09e78a43f2b57 is enough to restore proper behavior.

sergv:
@Bodigrim Looks like reverting https://github.com/UnkindPartition/tasty/commit/b152a0bc63166a4592e1f3639ef09e78a43f2b57 is the way to go: issuing `\r` clears part of line between first column and current cursor, but keeps content after cursor unchanged.

Because of printed progress message there may be text after cursor position which will not get cleared in ANSI-compliant terminals. Emacs seems to be not compliant and clears whole line when it sees `\r` so reverting the commit would not interfere with original intention.

PR #422, opened 2024-06-23
Don't depend on unbounded-delays on s390x

cboltz:
... which is a 64 bit architecture.

Note: s390x is big endian, but that hopefully doesn't matter.


Bodigrim (approved):
I verified that `cabal-3.10.2.1 check` (which is the version of Cabal currently used by Hackage) does not complain about `arch(s390x)`, so this is good to go.

Bodigrim:
Thanks!

Issue #424, opened 2024-07-17
tasty-quickcheck: a way to print out seeds for successful tests


phadej:
My property tests succeed, but some of the time they take unreasonably long time.

It would be great if it was possible to see the seeds used, so I can make a regressions.

---

Specifying timeout doesn't work, then `tasty` prints just

```
        LicenseExpression':                             TIMEOUT (0.04s)
          Timed out after 0.01
          Use -p '/LicenseExpression'\''/' to rerun this test only.
```

PR #425, opened 2024-07-17
Implement --quickcheck-timeout for individual tests within a property

### LINKED ISSUE #359: Output quickcheck seed on timeout
We had a quickcheck test timeout (using tasty's `--timeout` option), but we can't repro it because we don't have the seed. If we had the seed, we could see what quickcheck generated and see if it generated a pathological input that we haven't handled.

One possible solution that doesn't break the public API:
1. Pass the handle to the test thread (`asy`) to `applyTimeout`
2. Throw a new `TastyTimeout` exception to the test thread if the test thread takes too long
3. If `asy` returns a result within 1 second (or some other short time), return that result
4. Otherwise, do the current behavior: cancel `asy` and return a generic "Timed out" result
5. Update `tasty-quickcheck`'s `run` implementation to catch `TastyTimeout` and return a `Result` with the seed added to the description
    * Break out the current `timeoutResult` definition into `mkTimeoutResult :: TastyTimeout -> Result`

A possible solution that changes the `IsTest` interface (I know maintainers don't want to make breaking changes, but I think this would make `tasty` more flexible that can enable other future improvements):
* Add a new `TestMeta` type containing hooks for test providers; currently, it would contain a single field: `resultOnTimeout :: Maybe Result`
    * Pass an `IORef TestMeta` parameter to `run` that `run` can modify as it executes
    * Change `run :: test -> IO Result` to `run :: test -> IO (TestMeta, IO Result)`

---

<details>
<summary>Old solution that doesn't work</summary>

1. Add a new `TastyTimeout` exception containing information of the timeout, which should be thrown instead of the exception thrown by `timeout`
2. Break out `timeoutResult` into `mkTimeoutResult :: TastyTimeout -> Result`
7. Wrap the `tasty-quickcheck` [runner](https://github.com/UnkindPartition/tasty/blob/7379d3101f165a582aa3341084fbeaae9828875e/quickcheck/Test/Tasty/QuickCheck.hs#L214-L227) with

    ```hs
    handle (pure . mkTimeoutResultWithSeed replaySeed) $ do
      ...
    where
      -- N.B. technically violates the rule to always rethrow async exceptions,
      -- but i think this exception to the rule makes sense
      mkTimeoutResultWithSeed seed e =
        let result = mkTimeoutResult e
         in result{resultDescription = resultDescription result ++ printf " -- with seed: %d" seed}
    ```

This doesn't work because the test is run in a separate thread from the driver, and the timeout exception is handled in the driver thread and doesn't make it to the test thread. The test thread does get the `AsyncCancelled` exception thrown to it, but the driver thread never looks at the result of the test thread after the timeout, so it'll just get thrown away.
</details>

Bodigrim:
Presumably closes #359 and #424. 

@phadej @brandonchinn178 could you please give it a try?

### DIFF
--- a/quickcheck/tasty-quickcheck.cabal
+++ b/quickcheck/tasty-quickcheck.cabal
@@ -29,7 +29,7 @@ library
   other-extensions:    GeneralizedNewtypeDeriving, DeriveDataTypeable
   build-depends:       base >= 4.8 && < 5,
                        tagged < 0.9,
-                       tasty >= 1.5 && < 1.6,
+                       tasty >= 1.5.1 && < 1.6,
                        random < 1.3,
                        QuickCheck >= 2.10 && < 2.16,
                        optparse-applicative < 0.19
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -169,9 +169,10 @@ buildTestOutput opts tree =
                         ("",  pct) -> printf "%.0f%% " pct
                         (txt, 0.0) -> printf "%s" txt
                         (txt, pct) -> printf "%s: %.0f%% " txt pct
+              putChar '\r'
               -- A new progress message may be shorter than the previous one
               -- so we must clean whole line and print anew.
-              putChar '\r'
+              clearLine
               putStr testNamePadded
               infoOk msg
               hFlush stdout
@@ -190,6 +191,7 @@ buildTestOutput opts tree =
 
           when getAnsiTricks $ do
             putChar '\r'
+            clearLine
             putStr testNamePadded
 
           printFn (resultShortDescription result)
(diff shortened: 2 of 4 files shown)


brandonchinn178:
Looks great! Just ran it on a toy project

quickcheck/tasty-quickcheck.cabal
```diff
@@ -29,7 +29,7 @@ library
   other-extensions:    GeneralizedNewtypeDeriving, DeriveDataTypeable
   build-depends:       base >= 4.8 && < 5,
                        tagged < 0.9,
-                       tasty >= 1.5 && < 1.6,
+                       tasty >= 1.5.1 && < 1.6,
```
phadej, line 32: The change does what I need, thanks!

... but

The `tasty-1.5.1` on Hackage has `base <0` revision made.

If I use both `core` and `quickcheck` from the repo, then `cabal-install` finds an install plan.

Bodigrim: That's expected, I'll release `tasty-1.5.1.1` (or `1.5.2`) soon.

andreasabel (approved):

PR #427, opened 2024-08-13
Initial options

phadej:
It turns out `tasty` very much had all building blocks to allow specifying initial option set.

This PR is a draft, I'd welcome if anyone takes over as I did the least possible as a proof-of-concept (no haddocks for example).

The `options-example` "test-suite" is a demo.

`cabal run options-example -- --help ` prints

```
  --quickcheck-tests NUMBER
                           Number of test cases for QuickCheck to generate.
                           Underscores accepted: e.g. 10_000_000 (default: 1000)
```

(I made default value visible).

and if you run test the output is

```
options-example
  assoc: OK
    +++ OK, passed 1000 tests.
```

so the value is actually used. But it can still be overridden with `--quickcheck-tests`: `cabal run options-example -- --quickcheck-tests 1234`

```
options-example
  assoc: OK
    +++ OK, passed 1234 tests.
```

### DIFF
--- a/core/Test/Tasty.hs
+++ b/core/Test/Tasty.hs
@@ -36,7 +36,9 @@ module Test.Tasty
   , sequentialTestGroup
   -- * Running tests
   , defaultMain
+  , defaultMainWithOptions
   , defaultMainWithIngredients
+  , defaultMainWithIngredientsAndOptions
   , defaultIngredients
   , includingOptions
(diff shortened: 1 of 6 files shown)


Bodigrim:
This looks a reasonable way to address the problem in #414, but I'm extremely cautious to extend API beyond absolutely necessary. I'd like to hear from other maintainers.

PR #433, opened 2024-11-13
Add line wrap disablement to ConsoleReporter

edmundnoble:
I don't propose changing cabal.project, this is just for a demo until the PR to ansi-terminal is merged (https://github.com/UnkindPartition/ansi-terminal/pull/171/)

I saw that our test suite leaves garbage in the terminal output despite using `--hide-successes`. This turns out to be worse the narrower my terminal is. The more general problem is that `ConsoleReporter` leaves garbage in the terminal when the lines it prints are wrapped, because it doesn't know to clear multiple lines, just the most recent one. This PR fixes that by disabling line wrapping while printing that line.

This PR makes `ConsoleReporter`'s ANSI tricks clean up after themselves properly in the presence of a narrow terminal, a long test name, or a long progress message, including those printed by `testCaseWithSteps` from tasty-hunit. It also prevents test group names from wrapping, because that has the same cause.

### DIFF
--- a/core/Test/Tasty/Ingredients/ConsoleReporter.hs
+++ b/core/Test/Tasty/Ingredients/ConsoleReporter.hs
@@ -135,7 +135,6 @@ buildTestOutput opts tree =
     !alignment = computeAlignment opts tree
 
     MinDurationToReport{minDurationMicros} = lookupOption opts
-    AnsiTricks{getAnsiTricks} = lookupOption opts
 
     runSingleTest
       :: (IsTest t, ?colors :: Bool)
@@ -153,7 +152,7 @@ buildTestOutput opts tree =
           (replicate postNamePadding ' ')
 
         printTestName = do
-          putStr testNamePadded
+          withoutLineWrap $ putStr testNamePadded
           hFlush stdout
 
         printTestProgress progress
@@ -173,8 +172,9 @@ buildTestOutput opts tree =
               -- A new progress message may be shorter than the previous one
               -- so we must clean whole line and print anew.
               clearLine
-              putStr testNamePadded
-              infoOk msg
+              withoutLineWrap $ do
+                putStr testNamePadded
+                infoOk msg
               hFlush stdout
 
         printTestResult result = do
@@ -189,10 +189,11 @@ buildTestOutput opts tree =
                 _ -> fail
             time = resultTime result
 
-          when getAnsiTricks $ do
-            putChar '\r'
-            clearLine
-            putStr testNamePadded
+          withoutLineWrap $ do
+            when getAnsiTricks $ do
+              putChar '\r'
+              clearLine
+              putStr testNamePadded
 
           printFn (resultShortDescription result)
           when (floor (time * 1e6) >= minDurationMicros) $
@@ -211,7 +212,7 @@ buildTestOutput opts tree =
     runGroup _opts name grp = Ap $ do
       level <- ask
       let
-        printHeading = printf "%s%s\n" (indent level) name
+        printHeading = withoutLineWrap $ printf "%s%s\n" (indent level) name
         printBody = runReader (getApp (mconcat grp)) (level + 1)
       return $ PrintHeading name printHeading printBody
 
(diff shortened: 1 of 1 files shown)


core/Test/Tasty/Ingredients/ConsoleReporter.hs
```diff
@@ -223,6 +224,19 @@ buildTestOutput opts tree =
           , foldGroup = runGroup
           }
           opts tree
+  where
+    AnsiTricks{getAnsiTricks} = lookupOption opts
+    -- We must ensure these lines don't wrap, otherwise the wrong
+    -- line will be cleared later or the test tree printing will
+    -- itself wrap.
+    withoutLineWrap :: IO () -> IO ()
+    withoutLineWrap m = do
```
Bodigrim, line 233: Shall we use [`bracket`](https://hackage.haskell.org/package/base-4.20.0.1/docs/Control-Exception.html#v:bracket_)?

Bodigrim (commented):
That's a nice idea! Let's wait for https://github.com/UnkindPartition/ansi-terminal/pull/171 to be released, then I'll prepare a more in-depth review.

core/Test/Tasty/Ingredients/ConsoleReporter.hs
edmundnoble: I don't know really what the best practice is here. Printing is quick so I don't see it being interrupted by an async exception. Printing could result in an exception caused by stdout becoming unavailable, but in that case we can't print the `enableLineWrap` code anyway, so there is no recovery.

It does seem natural to use `bracket` though. It's up to you.

edmundnoble: Did it for now.

Bodigrim:
Please update the Cabal file with `ansi-terminal >= 1.1.2`.

Bodigrim:
Or, even better, let's guard usage of `disableLineWrap` / `enableLineWrap` by `#if MIN_VERSION_ansi_terminal(1,1,2)` and use no-op instead otherwise.

core/Test/Tasty/Ingredients/ConsoleReporter.hs
```diff
@@ -192,7 +192,8 @@ buildTestOutput opts tree =
           when getAnsiTricks $ do
             putChar '\r'
             clearLine
-            putStr testNamePadded
+            withoutLineWrap $
+              putStr testNamePadded
```
Bodigrim, line 196: What's the behavior of switching line wrapping off and on before finishing the line? We are going to print `resultShortDescription` later on.

edmundnoble:
@Bodigrim ansi-terminal 1.1.2 has been released and I've updated this PR to have it as a minimum bound. :partying_face:

core/Test/Tasty/Ingredients/ConsoleReporter.hs
edmundnoble: I'll investigate this as well as guarding the line wrap changes with CPP.

edmundnoble: I think we want `resultShortDescription` to also be printed without line wrap, so I'll put that in here.

edmundnoble:
The latest update should address all feedback thus far.

Bodigrim (commented):
We certainly benefit from no line wrapping when executing tests with `--hide-successes`. But do we benefit from disabling line wrapping in the absence of `--hide-successes`? Maybe it should be used only when `--hide-successes` is enabled?..

edmundnoble:
> We certainly benefit from no line wrapping when executing tests with `--hide-successes`. But do we benefit from disabling line wrapping in the absence of `--hide-successes`? Maybe it should be used only when `--hide-successes` is enabled?..

I argue that we do benefit from disabling line wrapping without `--hide-successes`: when printing test group names, we keep indenting further to the right, and on a narrow terminal, this may cause wrapping for longer test names.

Bodigrim:
> I argue that we do benefit from disabling line wrapping without `--hide-successes`: progress indicators get cleared even without --hide-successes, so this bug remains, especially using `testCaseWithSteps`.

That's a good argument.

My main concern is that without line wrapping `resultShortDescription` will be invisible on a narrow terminal. Which is especially undesirable if the result is `FAIL`. I think we should switch line wrapping on before printing `resultShortDescription` always or at least in case it is `FAIL`.

core/Test/Tasty/Ingredients/ConsoleReporter.hs
edmundnoble: I've verified that in at least my terminal, if I disable line wrap, then print enough text to ordinarily wrap, then enable line wrap and print some more text, the additional text after line wrap is enabled is wrapped and nothing else. I think that's what we want.

edmundnoble:
> > I argue that we do benefit from disabling line wrapping without `--hide-successes`: progress indicators get cleared even without --hide-successes, so this bug remains, especially using `testCaseWithSteps`.
> 
> That's a good argument.
> 
> My main concern is that without line wrapping `resultShortDescription` will be invisible on a narrow terminal. Which is especially undesirable if the result is `FAIL`. I think we should switch line wrapping on before printing `resultShortDescription` always or at least in case it is `FAIL`.

That's an excellent point. I undid that part of the change, now the result is not wrapped.

Bodigrim (approved):

Bodigrim:
Thanks, that's awesome!

PR #435, opened 2024-11-25
Tests: switch from pcre-light to regex-tdfa

Bodigrim:
`pcre-light` is not that light, because on a clean machine it fails with
```
Base.hsc:105:10: fatal error: 'pcre.h' file not found
#include <pcre.h>
         ^~~~~~~~
1 error generated.
```

There is no point to complicate build requirements, we can just use `regex-tdfa` instead.

### DIFF
--- a/quickcheck/tasty-quickcheck.cabal
+++ b/quickcheck/tasty-quickcheck.cabal
@@ -51,10 +51,10 @@ test-suite test
     test.hs
   build-depends:
       base >= 4.7 && < 5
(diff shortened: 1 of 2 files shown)


Bodigrim:
@andreasabel @VictorCMiraldo @martijnbastiaan just a gentle ping to review.

martijnbastiaan (approved):
Seems uncontroversial to me :)

andreasabel (commented):
Thumbs up!

PR #436, opened 2024-11-25
ConsoleReporter: force flushing of stdout after showCursor

Bodigrim:
I don't think I can write a test for this, but I have a persistent problem with some massive test suites when killing application with Ctrl+C leaves a terminal cursor hidden. I verified manually that flushing stdout fixes the issue.


VictorCMiraldo (approved):
Even if there's no tests, I don't think a call to `hFlush stdout` would be harmful here anyway! :+1:

PR #437, opened 2025-01-04
Release tasty-1.5.3, tasty-quickcheck-0.11.1

Bodigrim:
Release time! I'm keen to fix #436, which bugs me every time.


quickcheck/tasty-quickcheck.cabal
```diff
@@ -25,7 +21,6 @@ Source-repository head
 
 library
   exposed-modules:     Test.Tasty.QuickCheck
-  -- other-modules:
   other-extensions:    GeneralizedNewtypeDeriving, DeriveDataTypeable
   build-depends:       base >= 4.8 && < 5,
```
andreasabel, line 25: Would it make sense to lift `base` in this package to `>= 4.9`, since `tasty` has the same lower bound?

Bodigrim: Good idea, done.

andreasabel (approved):

PR #451, opened 2025-07-26
ConsoleReporter: limit max line length

### LINKED ISSUE #342: Add option to set max line length

Bodigrim:
Fixes #342 according with a plan in https://github.com/UnkindPartition/tasty/issues/342#issuecomment-1657088957

One thing that bothers me is that `buildTestOutput` is exposed, so adding `terminalWidth` as an argument technically constitutes a breaking change. Dunno, we can make it into a top-level binding:

```haskell
terminalWidth :: Maybe Int 
terminalWidth = fmap snd $ unsafePerformIO getTerminalSize
-- TODO this is a hack to avoid changing the type of `buildTestOutput`, remove it in the next major release
```

CC @brandonchinn178


Bodigrim:
Ci failures are because of https://github.com/UnkindPartition/ansi-terminal/issues/178

andreasabel:
I think this is a good feature!

I suppose there is no automatic testing of this feature because it does not do anything if the output is redirected, so we cannot have golden value tests for this change of formatting.
What is the status of manual testing?

Correction: The one thing that is verified by the golden values is:
> it does not do anything if the output is redirected

Bodigrim:
For manual testing one can run 
```
$ (cd core-tests && cabal run tasty-core-tests)
```
with different terminal width. A sample output on a narrow terminal looks like this:
```
Tests
  Resources
    Normal; a test excluded by a pattern:                  OK (0.10s)
    Exception during resource initialization:              OK
    Exception in test body; resource is released:          OK
    Exception in finalizer:                                OK
    Order of finalizers when the test suite runs:          OK (0.11s)
    Order of finalizers when the test suite is aborted (1 thread): OK (0.04s)
    Order of finalizers when the test suite is aborted (2 threads): OK (0.04s)
  Timeouts:                                                OK (0.20s)
  Dependencies
    True  AllSucceed:                                      OK (2.30s)
    True  AllFinish:                                       OK (2.30s)
    False AllSucceed:                                      OK (2.30s)
    False AllFinish:                                       OK (2.30s)
    show DependencyLoop:                                   OK
    Circular dependencies 1:                               OK
    Circular dependencies 2:                               OK
    Resource+dependencies interaction:                     OK (1.20s)
```
I also tested manually that progress reporting is not affected by this change.

PR #454, opened 2025-09-19
Separate modules in Cabal files with newlines (not commas)

Bodigrim:
Currently it's inconsistent: `tasty.cabal` uses both newlines and commas in the same fields. Let's use newlines uniformly.


Bodigrim:
(This was encountered while trying to build `tasty` with [MicroCabal](https://github.com/augustss/MicroCabal), apparently its parser is less permissive)

PR #457, opened 2025-10-02
Make tasty actually parallel by default

### LINKED ISSUE #455: NumThreads doesn't necessarily run tests in parallel?
hi, it seems that the `-j` option, backed by the [`NumThreads`](https://hackage-content.haskell.org/package/tasty/docs/Test-Tasty-Runners.html#t:NumThreads) type, is actually handled by [`runInParallel`](https://github.com/UnkindPartition/tasty/blob/master/core/Test/Tasty/Parallel.hs#L26), which only spawns the specified number of haskell worker threads, but without actually expanding the rts capability count via [`setNumCapabilities`](https://hackage.haskell.org/package/base/docs/Control-Concurrent.html#v:setNumCapabilities).

is this intended behavior? to me it's slightly surprising, since when i pass `-j8` and if the testsuite is linked with `-threaded`, then i would expect no less than 8 rts capabilities to be activated by `tasty` so that if the test cases are compute bound in haskell, there could be real parallelism to be harvested. the current behavior of `-j` doesn't do that, and would require user to also use `-with-rtsopts=-N` . if this is intended behavior that's also fine, though it's better if it's explicitly documented in `NumThreads` haddock documentation and `--help` output.

TerrorJack:
Closes https://github.com/UnkindPartition/tasty/issues/455.

### DIFF
--- a/core/Test/Tasty/Options/Core.hs
+++ b/core/Test/Tasty/Options/Core.hs
@@ -20,6 +20,8 @@ import GHC.Conc
 #if !MIN_VERSION_base(4,11,0)
 import Data.Monoid
 #endif
+import Control.Concurrent
+import System.IO.Unsafe
 
 import Test.Tasty.Options
 import Test.Tasty.Patterns
@@ -37,12 +39,13 @@ import Test.Tasty.Patterns
 newtype NumThreads = NumThreads { getNumThreads :: Int }
   deriving (Eq, Ord, Num)
 instance IsOption NumThreads where
-  defaultValue = NumThreads numCapabilities
+  defaultValue = unsafePerformIO $ NumThreads <$>
+      if rtsSupportsBoundThreads then getNumProcessors else pure 1
   parseValue = mfilter onlyPositive . fmap NumThreads . safeRead
   optionName = return "num-threads"
   optionHelp = return "Number of threads to use for tests execution"
   optionCLParser = mkOptionCLParser (short 'j' <> metavar "NUMBER")
-  showDefaultValue _ = Just "# of cores/capabilities"
+  showDefaultValue _ = Just "Number of cores when using threaded RTS, 1 for non-threaded"
 
 -- | Filtering function to prevent non-positive number of threads
 onlyPositive :: NumThreads -> Bool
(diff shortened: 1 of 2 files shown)


Bodigrim (approved):

andreasabel (approved):

Bodigrim:
Thanks!

arybczak:
I don't understand why this made it into the minor version bump, it's a significant change of behavior and also quite confusing, because now tasty completely ignores and overrides the number of capabilities set by the user with `+RTS -N`.

It's also worth noting that setting capabilities by default to the result of `getNumProcessors` results in an invalid behavior in a docker container with soft resource limits, because `getNumProcessors` returns the number of cores of the host.

Bodigrim:
Yeah, I didn't properly realise the consequences of the change, when I merged it. In the hindsight I should have bumped the major version. 

I'd imagine that oversubscribing threads to `getNumProcessors` inside a docker container is not too bad in practice, just a tad slower for RTS to manage. Or is it worse than that?

arybczak:
Let me elaborate a little. There are actually multiple problems with this PR:

1. It ignores `+RTS -Nx` setting (or whatever user sets with `setNumCapabilities` before running main tasty function), which is mostly confusing, but also somewhat limiting. I guess perhaps you can explicitly set NumThreads before running main tasty function in the code? If so, this takes care of the second part, but first remains.
2. It constrains you to the 1-1 relation of "running tests - capabilities". It's perfectly reasonable to want to run let's say 5 tests in parallel on 2 capabilities if they are somewhat I/O bound. Now you can't.
3. In a container with CPU limits it will by default try using all cores of the underlying host, which will be slower if it's a hard limit (because of redundant capabilities), but if it's a soft limit (the `resources` vs `limits` distinction in kubernetes) it might in addition slow down other containers currently running on the same host by taking their resources.
4. On bare metal on modern processors setting capabilities to `getNumProcessors` isn't actually a good default, because it reports the number of logical cores. So e.g. on Ryzen CPUs it will run twice as many capabilities as physical cores, which slows things down because each capability increases synchronization overhead which is far bigger than marginal gains from using more logical cores. Not sure how it works on Intel, might be slightly better but looks like they also have more logical cores than physical.

Because of these I suggest this PR is reverted. Status quo was conservative, but it didn't exhibit any of these issues.

Bodigrim:
Ignoring explicit `+RTS -Nx` is unfortunate; perhaps one can check `GHC.Environment.getFullArgs` and detect it?..

Not sure what would be a good way to aleviate the rest of the issues you listed. 

@TerrorJack what do you think?

PR #460, opened 2025-10-19
Don't depend on unbounded-delays for loongarch64

wszqkzqk:
LoongArch64 should fall into the same case in UnkindPartition#344 and should be added here. Tested locally.


Bodigrim (approved):

andreasabel (approved):

Bodigrim:
Thanks!

PR #466, opened 2026-02-03
Add instance IsTest (ContT () IO t)

Bodigrim:
Following up on our discussion with @mrkline at https://github.com/Bodigrim/tasty-bench/pull/69, how do we feel about adding `instance IsTest t => IsTest (ContT () IO t)`? Essentially it provides a way to run a test with a resource, which is
* more lightweight than [`withResource`](https://hackage-content.haskell.org/package/tasty-1.5.3/docs/Test-Tasty.html#v:withResource), which operates over `TestTree`s and is great to share a resource between tests but not that great if you want a single test,
* allows to deal with a more restrictive resource-providing API (imagine that a certain resource is provided only by `withFoo` combinator instead of having both `openFoo` and `closeFoo` exposed).

The new instance does not cost us anything in terms of dependencies. Any opinions?

### DIFF
--- a/core/Test/Tasty/Core.hs
+++ b/core/Test/Tasty/Core.hs
@@ -1,8 +1,10 @@
 -- | Core types and definitions
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE FlexibleInstances #-}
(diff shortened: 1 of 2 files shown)


andreasabel (approved):
I don't have any opinion on this and don't see how I could form one quickly.

PR #472, opened 2026-03-22
Fix display of HasCallStack backtraces when a test throws an error

Bodigrim:
Closes #469. There are two parts:

* Catch and rethrow exceptions with context in our vendored-in fork of `async`. This is inspired by upstream PR https://github.com/simonmar/async/pull/165.
* Bridge over multiple changes to `displayException` in `base` to ensure that at least some stack trace is shown.

### DIFF
--- a/core/Control/Concurrent/Async.hs
+++ b/core/Control/Concurrent/Async.hs
@@ -36,7 +36,7 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 -}
 
-{-# LANGUAGE MagicHash, UnboxedTuples #-}
+{-# LANGUAGE CPP, MagicHash, UnboxedTuples #-}
 
 module Control.Concurrent.Async (
   async, withAsync, wait, asyncThreadId, cancel, concurrently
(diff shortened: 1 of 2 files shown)


Bodigrim:
How to test this patch? Create a test, which throws an `error`, and observe `tasty` output. Here is a sample:

With GHC < 9.10:
```
    Foo: FAIL
      Exception: foo
      CallStack (from HasCallStack):
        error, called at Dependencies.hs:27:10 in core-tests-0.1-inplace-tasty-core-tests:Dependencies
        myTest, called at Dependencies.hs:20:20 in core-tests-0.1-inplace-tasty-core-tests:Dependencies
```

With GHC 9.10 (lots of garbage, but at least the first stack trace is relevant):
```
    Foo: FAIL
      Exception: foo
      CallStack (from HasCallStack):
        error, called at Dependencies.hs:27:10 in core-tests-0.1-inplace-tasty-core-tests:Dependencies
        myTest, called at Dependencies.hs:20:20 in core-tests-0.1-inplace-tasty-core-tests:Dependencies
      HasCallStack backtrace:
        collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:169:13 in ghc-internal:GHC.Internal.Exception
        toExceptionWithBacktrace, called at libraries/ghc-internal/src/GHC/Internal/IO.hs:260:11 in ghc-internal:GHC.Internal.IO
        throwIO, called at ./Control/Concurrent/Async.hs:122:7 in tasty-1.5.3-inplace:Control.Concurrent.Async
```

With GHC 9.12:
```
    Foo: FAIL
      Exception: foo

      HasCallStack backtrace:
        error, called at Dependencies.hs:27:10 in core-tests-0.1-inplace-tasty-core-tests:Dependencies
        myTest, called at Dependencies.hs:20:20 in core-tests-0.1-inplace-tasty-core-tests:Dependencies
```

With GHC 9.14:
```
    Foo: FAIL
      Exception: ghc-internal:GHC.Internal.Exception.ErrorCall:

      foo

      HasCallStack backtrace:
        error, called at Dependencies.hs:27:10 in core-tests-0.1-inplace-tasty-core-tests:Dependencies
        myTest, called at Dependencies.hs:20:20 in core-tests-0.1-inplace-tasty-core-tests:Dependencies
```

I was not aiming to match the whole message between GHCs, because it would require much more CPP. But at least all of versions point to `myTest` instead of omitting stack traces entirely or interrupting them inside of `Control.Concurrent.Async`.

core/Control/Concurrent/Async.hs
```diff
@@ -51,6 +51,12 @@ import GHC.Conc (ThreadId(..))
 import GHC.Exts
 import GHC.IO hiding (onException)
 
+#if MIN_VERSION_base(4,21,0)
+#define ExcWithContext(x) ExceptionWithContext (x)
+#else
+#define ExcWithContext(x) x
+#endif
```
andreasabel, line 79: Why isn't this a type synonym?
I think preference should be given to the capabilities of the Haskell language.  Only if you cannot express it in Haskell, revert to `#define`.
If it cannot be expressed by a Haskell type synonym, please add a comment explaining why this `#define` is necessary.

andreasabel (commented):
Is it possible to add a golden value test for this fix?

core/Control/Concurrent/Async.hs
Bodigrim: Fixed, thanks.

Bodigrim:
> Is it possible to add a golden value test for this fix?

It's hard to do so: we are testing Tasty's output, so a golden test would have to run another instance of Tasty to collect its output. Then, as demonstated [above](https://github.com/UnkindPartition/tasty/pull/472#issuecomment-4106182995), the output somewhat varies from version to version of GHC and contains highly volatile data (like exact line and column numbers).

andreasabel (approved):

PR #477, opened 2026-06-07
Use terminal-size package for terminal width detection

sergv:
The `terminal-size` package was proposed before. Current approach uses `ansi-terminal` for width detection which uses ANSI escape sequences by sending cursor repositioning request to stdout and reading resulting cursor position from stdin. In some setups, e.g. bash running under Emacs, using escape sequences like that indroduces unavoidable 0.5 seconds of delay which is very distruptive. The 'terminal-size` package uses system call to get terminal size and it doesn't introduce any delays.

Old approach of getting terminal size through `ansi-tricks` is kept around for cases where dependency footprint of `terminal-size` is undesirable (e.g. when running tests of the `bytestring` package).

Another change is to take `--ansi-tricks` flag into account and don't use ANSI escapes from `ansi-terminal` if ansi tricks are disabled.


Bodigrim:
Thanks!

PR #483, opened 2026-07-29
Better strategy for determining the default of NumThreads

arybczak:
Based on #482, but fixes its shortcomings (takes into account `-with-rtsopts`, `-maxN` and `--RTS` as well as `GHCRTS` env variable) and updates the README with an example how you can force sequential test runs from the code.


arybczak:
To be honest, I'd still put the default at cores/2 instead of cores because of point 4 from https://github.com/UnkindPartition/tasty/pull/457#issuecomment-4774013693, but I'm not hard pressed about it.

arybczak:
@Bodigrim ping

arybczak:
@Bodigrim :wave:

Bodigrim:
Sorry, I'm super slow these days. I'll do my best to get back to this by the end of the week.

